PR #49, opened 2015-03-03
[RA-578] Get OpenMRS 2.2 running atop OpenMRS Platform 1.11

k-joseph:
added the ChartSearch module into the reference application distro module


k-joseph:
Pushed directly to master since i would tell from where the conflicts were showing up from

PR #53, opened 2015-06-10
RA-740 added cancel from review

tomaszmueller:


tomaszmueller:
https://issues.openmrs.org/browse/RA-740

rkorytkowski:
You should be able to set database.username, database.password, jdbc.connectionurl in https://github.com/openmrs/openmrs-distro-referenceapplication/blob/master/ui-tests/src/main/resources/org/openmrs/uitestframework/test.properties

The class which reads that is https://github.com/openmrs/openmrs-contrib-uitestframework/blob/master/src/main/java/org/openmrs/uitestframework/page/TestProperties.java

Please add that info to https://wiki.openmrs.org/display/docs/UITestFramework if it works

PR #56, opened 2015-06-17
RA-788 Automating RA-718 Active Visit Edit Patient Test

nataliaplonka:
https://issues.openmrs.org/browse/RA-788

### DIFF
--- a/package/pom.xml
+++ b/package/pom.xml
@@ -4,7 +4,7 @@
 
 	<groupId>org.openmrs.distro</groupId>
 	<artifactId>referenceapplication-package</artifactId>
-	<version>2.2-SNAPSHOT</version>
+	<version>2.2</version>
 	<packaging>pom</packaging>
 	<name>Reference Application Distribution Package</name>
 	<description>Scripts for building the Reference Application distribution package</description>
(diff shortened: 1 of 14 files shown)


tomaszmueller:
@nataliaplonka are the methods: isElementPresent, isAlertPresent, closeAlertAndGetItsText actually needed?

nataliaplonka:
No :) I improved it.

tomaszmueller:
@nataliaplonka:
I still don't see the usage of these methods in this test case. Remove them please.
However, these methods may be useful in other test cases, so, in your future automated tests, think if you can use them before removing.
Second thing: In: 'Assert.assertEquals("Today", patientDashboardPage.visitLink().getText().trim());' the assertion isn't needed. You should change it into waiting for the visitlink element.
If you pull the latest changes from upstream you should have   waitForVisitLink() function in patientDashboardPage you can use.
Third thing: In the test case steps you have patient named "John Bob Smith", while in the automation test you have "John   Bob". The gender field also doesn't seem to be filled-in(does the test work?). 

And please put assertions in at least the most important places in the test to check Expected Result.
Regards,
Tomasz Mueller

nataliaplonka:
I removed this method and assertion, you saw this when I commuted the changes I don't know why they aren't save.

tomaszmueller:
@nataliaplonka git sometimes does strange things and you have situations when you commit & push changes and findout they aren't there. The best way is to check in 'Files changed' tab, what changes have actually been pushed.

PR #64, opened 2015-07-06
Ra 816 Automating: RA-695 Add new Allergy Test; Ra-691 Edit Allegry Test; RA-704 Delete Allergy Test

nataliaplonka:
https://issues.openmrs.org/browse/RA-816


tomaszmueller:
@nataliaplonka
1. In AllergyPage, editPresent, you shouldn't use 'driver.findElement' if you can use just 'findElement'.
2. These tests need to be refactored so that added alergy is always deleted and the edition is always taking part on an added allergy, not allergy already existing

nataliaplonka:
If I added the same allergy to existing allergy then I get error message, that's why If allergy exist I edit or deleting this allergy but if doesn't exist, I'll create the new one. Why I can't use 'driver.findElement, it give me the same result?

dkayiwa:
@tomaszmueller do we still need this?

tomaszmueller:
@dkayiwa Yes, @nataliaplonka is still working on this issue

PR #66, opened 2015-07-07
RA-822 Automating: RA-699 Start Visit Test

nataliaplonka:
https://issues.openmrs.org/browse/RA-822


tomaszmueller:
@nataliaplonka
changes in ProviderTest should not be in this commit

PR #71, opened 2015-07-13
RA-829 Automating: RA-679 Register and Find Patient Record by ID or Name or Surname

nataliaplonka:
https://issues.openmrs.org/browse/RA-829


tomaszmueller:
@nataliaplonka 
import org.apache.commons.lang3.StringUtils; is not needed

nataliaplonka:
For the future please tell me which file, it would save me time in finding.

tomaszmueller:
You're right I should've told you which file is it, sorry...

PR #86, opened 2015-07-23
RA-819

yulia-kh:
Implemented test for improvement https://issues.openmrs.org/browse/RA-819 Automating RA-653 Find Patient Record


tomaszmueller:
@yulia-kh First of all, we have a rule to not to use classes imported from selenium(e.g: By, WebElement) directly in *Test classes. They should be moved to some functions(methods) in *Page classes and a *Test class should only use those functions.
Second, why did you use Thread.sleep instead of WebDriverWait that you have commented? What kind of problems did it cause? And why is sleep/wait needed in this place?

yulia-kh:
@tomaszmueller fixed all the issues:
selenium logic moved to *Page, and I've found the way how not to use Thread.sleep. 
Please check.

tomaszmueller:
@yulia-kh  Looks OK, just two little issues left:
in *Page files you can use findElement method directly, so please don't use driver.findElement(lines 58 and 60 of FindPatientPage.java)
You should put By.cssSelector(".dataTables_empty") into final field like you did with PATIENT_ID_SEARCH_RESULT.

yulia-kh:
Done

tomaszmueller:
@yulia-kh I meant rather creating a field like:
private static final By DATA_TABLES_EMPTY = By.cssSelector(".dataTables_empty");
and then using it like: findElement(DATA_TABLES_EMPTY)
but the way you did it is ok, too
I'm merging this

tomaszmueller:
Thanks, @yulia-kh!

PR #90, opened 2015-07-24
RA-881 Automating: RA-707 Using Back button in Merge Patient Electron…

nataliaplonka:
https://issues.openmrs.org/browse/RA-881


tomaszmueller:
@nataliaplonka this test actually wasn't suppose to test the merging itself, but the fact that 'Continue' button is disabled after putting just the first id of the patient.
It was created to check whether the issue : https://issues.openmrs.org/browse/RA-641 is fixed

PR #99, opened 2015-10-07
RA-819: Replace WebDriverWait by Thread.sleep() for waiting result table

yulia-kh:
Hi Tomasz,

I've updated the test, now it use Thread.sleep() with progressive timeout instead of WebDriverWait. But I still think that the previous version should work correct. May be 5 second timeout in line 
WebDriverWait wait = new WebDriverWait(driver, 5) 
was not enough.


tomaszmueller:
Thanks, @yulia-kh ! :)

PR #101, opened 2016-01-10
LICENSE CREATED

Basarqari:
License was missing,so I added


dkayiwa:
@Basarqari where do you get the license file from? It seems to be different from those of other modules.

PR #106, opened 2016-03-18
Update idgen to 4.3 RA-1084

cardy31:
devtest01 isn’t letting me run UI tests because it can’t boot up the
registration modules due to idgen not being up to date. This fixes that.


dkayiwa:
@cardy31 can you squash the two commits into one?

djazayeri:
I just committed this change directly. Thanks @cardy31

PR #109, opened 2016-04-07
Updated hostname of test server from devtest01 to int02

yancmo:


ui-tests/pom.xml
```diff
@@ -17,7 +17,7 @@
 
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-        <webapp.url>http://devtest01.openmrs.org:8080/openmrs</webapp.url>
+        <webapp.url>http://int02.openmrs.org:8080/openmrs</webapp.url>
```
djazayeri: Actually please use int-refapp, which is a clearer alias for our purposes. And we don't need the :8080 anymore.

yancmo:
I think we'll need to update [https://github.com/openmrs/openmrs-distro-referenceapplication/blob/master/utility-scripts/appStartupCheck.sh](appStartupCheck.sh) and [https://github.com/openmrs/openmrs-distro-referenceapplication/blob/master/utility-scripts/tomcatLogCollector.sh](tomcatLogCollector.sh) as well.. Should I update that?

dkayiwa:
@yancmo i already did. Thanks 👍

PR #110, opened 2016-04-08
RA-1102 Improved reliability of AddDiagnosisToVisitNoteTest

yancmo:
Ticket: https://issues.openmrs.org/browse/RA-1102
- Removed end visit at end of test so that the number of patients in active visits wont decrease
- Detect end visit link (which will always be there if page is returned successfully), rather than success toast

### DIFF
--- a/ui-tests/src/test/java/org/openmrs/reference/AddDiagnosisToVisitNoteTest.java
+++ b/ui-tests/src/test/java/org/openmrs/reference/AddDiagnosisToVisitNoteTest.java
@@ -13,7 +13,6 @@ import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNotNull;
 
 import org.junit.Test;
-import org.junit.Ignore;
 import org.junit.experimental.categories.Category;
 import org.openmrs.reference.groups.BuildTests;
 import org.openmrs.reference.page.ClinicianFacingPatientDashboardPage;
@@ -25,7 +24,6 @@ public class AddDiagnosisToVisitNoteTest extends ReferenceApplicationTestBase {
     private VisitNotePage visitNotePage;
     
     @Test
-    @Ignore //unstable
     @Category(BuildTests.class)
     public void AddDiagnosisToVisitNoteTest() throws Exception {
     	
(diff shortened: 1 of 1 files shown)


ui-tests/src/test/java/org/openmrs/reference/AddDiagnosisToVisitNoteTest.java
```diff
@@ -36,8 +34,6 @@ public void AddDiagnosisToVisitNoteTest() throws Exception {
         assertEquals("Pneumonia", visitNotePage.primaryDiagnosis());
         assertEquals("Bleeding", visitNotePage.secondaryDiagnosis());
         visitNotePage.save();
-        assertNotNull(patientDashboardPage.visitLink());
```
dkayiwa: Can you do some bit of explaining as to how this improves the reliability of this test?

yancmo: I was previously working on this test, which got merged after passing through the CI. However, it failed a few days later on the [CI](https://ci.openmrs.org/browse/REFAPP-OMODDISTRO-4441/log) at that line, even though it was working fine on my local machine.
Previously, what visitLink() does is to detect a success toast pop up after saving the note. Back then, the test failed as it could not detect it for some reason (maybe there was no popup after saving it?), and caused a timeout. So what the change does is to detect another web element in the page instead; whether the page is returned to successfully, is the main objective.

On a side note, it seems like the log for UI tests is not displaying correctly. It used to display the part with the full error stack, but now it's only displaying the bottom part of it.

dkayiwa:
@yancmo ok merged. Let us see if this time we do not fail randomly. :)

dkayiwa:
@yancmo it actually failed for the first time after merging as you can see here: https://ci.openmrs.org/browse/REFAPP-OMODDISTRO-4511
If it keeps failing, i will have no option but revert it. :)

yancmo:
Hmm, it seems like it's still failing at that same part of the code, not sure why as it's working fine locally. Was it running successfully at https://ci.openmrs.org/browse/REFAPP-OMODDISTRO-4516? If so what were the changes made? But yea, if it keeps failing I guess you can revert or disable it.

dkayiwa:
It has again failed two more times including https://ci.openmrs.org/browse/REFAPP-OMODDISTRO-INTTESTS-4522/log
So am reverting

PR #131, opened 2016-08-24
RA-1102 - Fixed reliability of AddDiagnosisToVisitNoteTest

tmarzeion:
This should work when max timeout will be increased in this PR:
https://github.com/openmrs/openmrs-contrib-uitestframework/pull/9

### DIFF
--- a/ui-tests/src/test/java/org/openmrs/reference/AddDiagnosisToVisitNoteTest.java
+++ b/ui-tests/src/test/java/org/openmrs/reference/AddDiagnosisToVisitNoteTest.java
@@ -25,7 +25,6 @@ public class AddDiagnosisToVisitNoteTest extends ReferenceApplicationTestBase {
     private VisitNotePage visitNotePage;
     
     @Test
-    @Ignore //unstable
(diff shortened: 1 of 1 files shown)


ui-tests/src/test/java/org/openmrs/reference/AddDiagnosisToVisitNoteTest.java
```diff
@@ -36,7 +35,7 @@ public void AddDiagnosisToVisitNoteTest() throws Exception {
         assertEquals("Pneumonia", visitNotePage.primaryDiagnosis());
         assertEquals("Bleeding", visitNotePage.secondaryDiagnosis());
         visitNotePage.save();
-        assertNotNull(patientDashboardPage.visitLink());
+        visitNotePage.waitForElement(ClinicianFacingPatientDashboardPage.VISIT_LINK);
```
rkorytkowski: visitNotePage.save() should return ClinicianFacingPatientDahboardPage and only ther you can look for elements. But looking for a toast message is bad. Look for entered diagnoses in the diagnoses section.

PR #139, opened 2016-08-25
RA-1191 Fix FindPatientRecordTest

AdamGrzybkowski:


ui-tests/src/main/java/org/openmrs/reference/page/FindPatientPage.java
```diff
@@ -13,7 +13,7 @@
 
     public static final By PATIENT_SEARCH_RESULT = By.id("patient-search-results-table");
     public static final By PATIENT_NAME_SEARCH_RESULT = By.xpath("//table[@id='patient-search-results-table']/tbody/tr/td[2]");
-    public static final By PATIENT_ID_SEARCH_RESULT = By.xpath("//table[@id='patient-search-results-table']/tbody/tr/td[1]");
+    public static final By PATIENT_ID_SEARCH_RESULT = By.cssSelector("#patient-search-results-table td");
```
rkorytkowski: Doesn't it select more than one td? How about "#patient-search-results-table tr:first-child td:first-child" See http://www.w3schools.com/cssref/css_selectors.asp ?

ui-tests/src/main/java/org/openmrs/reference/page/FindPatientPage.java
```diff
@@ -13,7 +13,7 @@
 
     public static final By PATIENT_SEARCH_RESULT = By.id("patient-search-results-table");
     public static final By PATIENT_NAME_SEARCH_RESULT = By.xpath("//table[@id='patient-search-results-table']/tbody/tr/td[2]");
```
rkorytkowski: It doesn't seem to be used anywhere...

ui-tests/src/main/java/org/openmrs/reference/page/FindPatientPage.java
```diff
@@ -13,7 +13,7 @@
 
     public static final By PATIENT_SEARCH_RESULT = By.id("patient-search-results-table");
```
rkorytkowski: Let's make all those locators private instead of public so that only this page knows how it looks like.

I'll add a note the the guide.

rkorytkowski:
Committed in https://github.com/openmrs/openmrs-distro-referenceapplication/commit/aa758e70e66f45ea96f653c353399eabd286947e

PR #153, opened 2016-08-31
RA-1203 fix CreateAppointmentBlockTest

AdamGrzybkowski:
@rkorytkowski two questions:
1) Do you have an idea how to get only appointments form specific day by e.g css selector? 
2) Is there a way to create service via REST ? We could create test service and use it instead the existing one.


rkorytkowski:
I don't have ideas off hand for 1). Don't spend to much time on that and if it is not easily doable just remove the test.

2) is most likely not possible unless you add a rest endpoint. It's not a priority so I wouldn't do that. Simply remove tests which need that.

AdamGrzybkowski:
So if we can't create service I would rather delete this test, because we can't really check if the appointment was created, we can search by service name, but it is not unique.

rkorytkowski:
@AdamGrzybkowski fair enough

ui-tests/src/main/java/org/openmrs/reference/page/AppointmentSchedulingPage.java
```diff
@@ -0,0 +1,35 @@
...
+ * Copyright (C) OpenMRS Inc. OpenMRS is a registered trademark and the OpenMRS
+ * graphic logo is a trademark of OpenMRS Inc.
+ */
+package org.openmrs.reference.page;
+
+import org.openmrs.uitestframework.page.Page;
+import org.openqa.selenium.By;
+
+public class AppointmentSchedulingPage extends Page{
+
```
AdamGrzybkowski: @rkorytkowski 
Should I leave this page ? Or just delete CreateAppointmentBlockTest ?

PR #159, opened 2016-09-01
RA-1210 - Fixed AddUserTest

tmarzeion:
Takes about 2 minutes because of `addEditUserPage.getValidationErrors()` loading time.
No idea why it takes so long, I've used this method before and It never took so long.

### DIFF
--- a/ui-tests/src/main/java/org/openmrs/reference/page/AdministrationPage.java
+++ b/ui-tests/src/main/java/org/openmrs/reference/page/AdministrationPage.java
@@ -11,32 +11,27 @@ package org.openmrs.reference.page;
 
 import org.openmrs.uitestframework.page.Page;
 import org.openqa.selenium.By;
-import org.openqa.selenium.WebDriver;
 
 public class AdministrationPage extends Page {
 
-    public static String URL_PATH = "/admin/index.htm";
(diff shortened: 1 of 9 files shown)


ui-tests/src/main/java/org/openmrs/reference/page/AdministrationPage.java
```diff
@@ -11,32 +11,27 @@
...
-    private static By MANAGE_USERS = By.cssSelector("#content  table:nth-child(4)  div:nth-child(1) li:nth-child(2)  a");
-    private final static By MANAGE_VISIT_TYPES = By.cssSelector("#content  table:nth-child(4)  div:nth-child(4) li:nth-child(2)  a");
-    private static By MANAGE_PROVIDERS = By.cssSelector("#content a[href='/openmrs/admin/provider/index.htm']");
+    private static final String URL = "/admin/index.htm";
 
-    public AdministrationPage(WebDriver driver) {
-        super(driver);
-    }
+    private static final By MANAGE_USERS = By.linkText("Manage Users");
+    private final static By MANAGE_VISIT_TYPES = By.cssSelector("#content  table:nth-child(4)  div:nth-child(4) li:nth-child(2)  a");
```
rkorytkowski: Add an id to the page. This selector seems unreliable.

PR #170, opened 2016-09-06
RA-1220 Fix creating test visit in UI tests

AdamGrzybkowski:


ui-tests/src/test/java/org/openmrs/reference/ReferenceApplicationTestBase.java
```diff
@@ -33,4 +35,8 @@ public void before() throws Exception {
 	    homePage = new HomePage(page);
 	}
 
+	protected String getLocationUuid(){
+		return homePage.findElement(SELECTED_LOCATION).getAttribute("location-uuid");
```
rkorytkowski: It won't work in all cases homePage.findElement will be waiting for loading the homepage before trying to locate the element. If you call this method on any other page then it will fail. I'd change it to public String getLocationUuid(Page page);

AdamGrzybkowski:
@rkorytkowski PR updated

PawelGutkowski:
Merged, thanks!

PR #174, opened 2016-09-07
RA-1225 fix MergeVisitsTest

AdamGrzybkowski:


ui-tests/src/main/java/org/openmrs/reference/page/ClinicianFacingPatientDashboardPage.java
```diff
@@ -325,12 +325,13 @@ public PatientVisitsDashboardPage addPastVisit(){
 	}
 
 
-	public void clickChangeDate (){
+	public PatientVisitsDashboardPage clickChangeDate (){
```
rkorytkowski: Does it really redirects to PatientVisitsDashboardPage?

PR #193, opened 2016-09-09
RA-1236 - Fixed CheckModulesTest fails on fresh distribution

tmarzeion:
This PR is dependent by: https://github.com/openmrs/openmrs-module-legacyui/pull/62

### DIFF
--- a/ui-tests/src/test/java/org/openmrs/reference/CheckModulesTest.java
+++ b/ui-tests/src/test/java/org/openmrs/reference/CheckModulesTest.java
@@ -7,32 +7,24 @@ import org.junit.Before;
 import org.junit.Ignore;
 import org.junit.Test;
 import org.junit.experimental.categories.Category;
+import org.openmrs.reference.page.AdministrationPage;
 import org.openmrs.reference.page.HomePage;
 import org.openmrs.reference.page.ModulesPage;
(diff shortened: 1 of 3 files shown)


ui-tests/src/test/java/org/openmrs/reference/CheckModulesTest.java
```diff
@@ -7,32 +7,24 @@
...
      */
     @Test
     @Category(org.openmrs.reference.groups.BuildTests.class)
-    @Ignore("Test doesn't extend ReferenceApplicationTestBase and fails on fresh distribution")
     public void checkModules() throws Exception {
-    	assertPage(homePage);
-        modulesPage.go();
+        AdministrationPage administrationPage = homePage.goToAdministration();
+        ModulesPage modulesPage = administrationPage.goToManageModulesPage();
         Thread.sleep(500);
```
rkorytkowski: The sleep should be removed. We don't want to rely on that.

ui-tests/src/test/java/org/openmrs/reference/CheckModulesTest.java
```diff
@@ -7,32 +7,24 @@
...
         assertPage(modulesPage);
```
rkorytkowski: The assert should be removed. We don't do that.

PR #194, opened 2016-09-09
RA-1237 - Fixed comeBackToApplicationAfterLogoutTest

tmarzeion:
This test is very tiny


ui-tests/src/test/java/org/openmrs/reference/ComeBackToApplicationAfterLogoutTest.java
```diff
@@ -1,43 +1,23 @@
...
-        
-        assertPage(homePage);
-    }
-
-    @After
-    public void tearDown ()throws Exception {
-        headerPage.clickOnHomeIcon();
-        headerPage.logOut();
+        //Logout
+        homePage.goToPage("/logout");
```
rkorytkowski: Do not hardcode /logout. Use getLoginPage().go();

PR #200, opened 2016-09-15
RA-1179 display server logs after UI tests on Travis CI

PawelGutkowski:


rkorytkowski:
Thanks! Let's see how it works.

PawelGutkowski:
When build fails it exceeds maximum time, so it seems that changes I introduced are useless...

PawelGutkowski:
There is a `after_failure` lifecycle hook, I will investigate if it gets executed even when timeout occurs

travis_after_script.sh
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env bash
+
+if [ "$TRAVIS_PULL_REQUEST" = "false" ]
+then
+docker exec openmrs-referenceapplication cat .OpenMRS/openmrs.log
```
rkorytkowski: Could we please echo "\n\nLoading .OpenMRS/openmrs.log from the server:\n\n" before printing out logs?

PawelGutkowski: Okay, I will add it in follow-up. After reading this https://docs.travis-ci.com/user/customizing-the-build#Build-Timeouts I think that Travis is killing our build after 50 minutes(with VM startup included) no matter what we put into config. 

Maybe we could find some way to detect if server started successfully and stop test suite much earlier If it failed? That would save our time, resources, and finally we would see server logs...

rkorytkowski: Okay. We could fail a test when the server responds with error HTTP status code when accessing the login page. The change would have to be implemented in uitestframework.

PawelGutkowski: @rkorytkowski all right, I will take care of it

PR #204, opened 2016-11-21
OCLM-50 - Add OCL module to Reference Application

tmarzeion:
This PR adds 1.2-SNAPSHOT version of OpenMRS OCL module to Refapp distro.
@rkorytkowski We should use 1.2 version here instead of SNAPSHOT, but we need to wait for release :)


dkayiwa:
@tmarzeion do you have an idea regarding what is happening here? https://ci.openmrs.org/browse/REFAPP-DUT-3544

PR #216, opened 2017-03-20
RA-1335: Updated readme - Vagrant marked outdated and added SDK

merovingienne:
Ticket: https://issues.openmrs.org/browse/RA-1335

Edit #1
Updated the Readme to reflect that the Vagrant box is no longer maintained, and added OpenMRS SDK as the method to locally build and run the Ref App.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,8 +7,20 @@ openmrs-distro-referenceapplication
 
 [![Build Status](https://saucelabs.com/browser-matrix/rkorytkowski.svg)](https://saucelabs.com/open_sauce/user/openmrs)
 
-## Running locally with Vagrant 
-https://wiki.openmrs.org/x/CIC3Ag
+## Running locally with OpenMRS SDK
+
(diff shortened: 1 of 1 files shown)


dkayiwa:
@merovingienne to ensure that we do not have the same documentation in two places, shouldn't this just have a link to the SDK wiki page?

merovingienne:
But that doesn't explain how to deploy a local build, does it? As I understood, the instructions there say how to run/ deploy a ref app that has been already built, and is kept in the repos. i.e. not the local build with our changes?

dkayiwa:
@merovingienne then in that case, what value does the SDK add?
By the way, have you looked at this? https://wiki.openmrs.org/display/docs/OpenMRS+Standalone

merovingienne:
The ability to separately develop multiple instances of the Reference app. 
eg: Work on an issue related to the Ref App, then test it on multiple platform versions with different servers made with the SDK. The SDK makes it easy to manage multiple server instances for development.

The other method to do development work and test them for the Reference application itself would be with Tomcat. It's much easier with the SDK to run multiple servers.

If this does not seem useful, I'll only add the link to the SDK.

dkayiwa:
In addition to the SDK, did you notice that the standalone can also let you run multiple server instances by just having each server as a separate folder?

merovingienne:
Oh yeah, just noticed that.

I'll put another commit with a simple reference to the SDK wiki page :)

PR #222, opened 2017-10-01
RA-1193: Test on July failed, tryed to re-enable it 

DomenicoDiLeo:
According to the [topic]((only the  ignore annotation has been removed)) , we have to re-enable this test and investigate whenever it fails on SauceLab

### DIFF
--- a/ui-tests/src/test/java/org/openmrs/reference/AddPastVisitTest.java
+++ b/ui-tests/src/test/java/org/openmrs/reference/AddPastVisitTest.java
@@ -11,7 +11,6 @@ package org.openmrs.reference;
 
 import org.junit.After;
 import org.junit.Before;
-import org.junit.Ignore;
 import org.junit.Test;
 import org.junit.experimental.categories.Category;
 import org.openmrs.reference.groups.BuildTests;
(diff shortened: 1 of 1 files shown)


dkayiwa:
@DomenicoDiLeo did you look at the travis failures?

DomenicoDiLeo:
Yes, I  noticed it, but I need some time to understand the error. It seems that something has changed in the developement process. In the past after a commit, the test case were built to run against to the online server https://qa-refapp.openmrs.org/openmrs. They failed because I didn't have a sauce lab account. Now they are failing because the test are executed against localhost, I get this error message:
org.openqa.selenium.firefox.NotConnectedException: Unable to connect to host localhost on port 7055.
Is there any server active on that port. 
Keep in mind when I run locally the tests, they are executed against https://qa-refapp.openmrs.org/openmrs and not on my local PC ( I have no openmrs server installed). Additionally, to my knowledge I can change the address of the server only in the [pom.xml ](https://github.com/DomenicoDiLeo/openmrs-distro-referenceapplication/blob/master/ui-tests/pom.xml). It is strange to me that the tests are executed against localhost:7055 after the commit. It sounds like  they are executed on the build server which should have an Openmrs locally installed. 
Are you aware of any change in the build process?
Thank you

DomenicoDiLeo:
@dkayiwa any news? I've also wrote a post on [opemrs](https://talk.openmrs.org/t/ui-test-uilding-failure/13662)

dkayiwa:
@DomenicoDiLeo i have responded from the talk thread.

PR #250, opened 2019-02-03
Removed logic module as a dependency

reubenvarghese1:
The logic module is not deprecated and is no longer part of the ref app distro. Hence all dependencies to it have been removed.


reubenvarghese1:
@dkayiwa , does this look good to you?

dkayiwa:
Oh yes. Thanks @reubenvarghese1

PR #251, opened 2019-02-03
RA-1571: Add Attachments module to Reference Application Distribution

reubenvarghese1:
From Ref App 2.9 onwards, the [attachments module](https://addons.openmrs.org/show/org.openmrs.module.attachments) will officially be a part of the distribution.

Issue: https://issues.openmrs.org/browse/RA-1571

PR #252, opened 2019-03-09
RA-1571: Add Attachments module to Reference Application Distribution(Follow-up commit)

reubenvarghese1:
RA-1571: Add Attachments module to Reference Application Distribution(Follow-up commit)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -314,6 +314,12 @@
 				<scope>provided</scope>
 			</dependency>
 			<dependency>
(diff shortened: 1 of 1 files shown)


reubenvarghese1:
@dkayiwa, does this look right to you?

This is being done as a follow-up commit to [this](https://github.com/openmrs/openmrs-distro-referenceapplication/pull/251#issuecomment-466763259) discussion

PR #257, opened 2019-09-11
RA-1592 : Boostrap as a foundation for Reference Application UI Project.

Akayeshmantha:
This PR contains changes for the GSoC Project Bootstrap as a foundation for Ref App UI project.
It contains the following changes up to now

1.Updated the modules to the latest snapshot.

This PR will have more commit's when the project is going forward.


Akayeshmantha:
@ssmusoke other than this two modules other modules were in the latest snapshot versions.

PR #259, opened 2019-10-03
RA-1627: Bundle the GrowthCharts Module to Reference Application Distribution

mozzy11:
Ticket :https://issues.openmrs.org/browse/RA-1627

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,7 +23,7 @@ before_cache:
  
 install: true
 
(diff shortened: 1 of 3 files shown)


dkayiwa:
I do not believe that we should bundle this module with the  reference application. It is not widely used.

mozzy11:
@dkayiwa , Alright . that makes sense. Whoever shall need it will deploy it them selves

PR #270, opened 2020-04-05
RA-1733:Upgrade Reference Aplication distro Module Versions

sherrif10:
Updated the reference application distro module version
 https://issues.openmrs.org/browse/RA-1733
  cc @mozzy


sherrif10:
Thanks @mozzy11

PR #273, opened 2020-09-04
RA-1818:Followup commit 

sherrif10:
Followup

PR #274, opened 2020-09-09
Revert "RA-1818:Followup commit "

mozzy11:
Reverts openmrs/openmrs-distro-referenceapplication#273

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -28,10 +28,9 @@
   </scm>
 
 	<properties>
-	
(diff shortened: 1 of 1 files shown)


mozzy11:
Reasons for this revert, still looking into details why the standalone build fails at https://ci.openmrs.org/browse/REFAPP-OMODDISTRO-TS-10273/log .

Its releated to liquibase upgrade in the core

PR #288, opened 2020-10-20
Updating Fhir2 version to 1.1.0-snapshot

sherrif10:
This version of fhir2 need to be released to  maintenance version. **WHY** Because there was a quick fix here https://github.com/openmrs/openmrs-module-fhir2/commit/19c306101dc9dcd56aa28988d31295d575ae6ebb ,So we needed to release this version to fix a problem in refApp distro
cc @mozzy11  @ibacher


mozzy11:
@sherrif10  , can you squash those commits ??
they seem to be irrelevant to the issue you're addressing here

PR #297, opened 2021-02-08
fix

mozzy11:


dkayiwa:
A commit message like this is fir for a /dev/-1

mozzy11:
haha :smile:   , @dkayiwa   am just creating a minor PR to fix the failure . i only simply have to create a PR , wait for it to build succesfully and it will fix the Bamboo failure

dkayiwa:
That does not justify such a commit message.

mozzy11:
so  the Build has passed ,as long as This PR builds succesful .i even dont have to merge it

dkayiwa:
Bamboo is based on only merged pull requests. There is no way it is affected by unmerged pull requests like this.

PR #300, opened 2021-02-27
Ratest 77: Added attachment attribute on the patient dashboard page

kdaud:
Here is the ticket id https://issues.openmrs.org/browse/RATEST-77
Added the attachment attribute on the patient dashboard page

### DIFF
--- a/ui-tests/src/test/java/org/openmrs/reference/page/ClinicianFacingPatientDashboardPage.java
+++ b/ui-tests/src/test/java/org/openmrs/reference/page/ClinicianFacingPatientDashboardPage.java
@@ -9,12 +9,12 @@
  */
 package org.openmrs.reference.page;
 
-import org.apache.commons.lang.StringUtils;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Calendar;
 import java.util.List;
 
+import org.apache.commons.lang.StringUtils;
 import org.openmrs.uitestframework.page.Page;
 import org.openqa.selenium.By;
 import org.openqa.selenium.WebElement;
@@ -85,9 +85,10 @@ public class ClinicianFacingPatientDashboardPage extends Page {
 
 	private static final By ALLERGIES_LINK = By.id("allergyui-editAllergies");
 
-	private static final By CONDITIONS_LINK = By
-			.cssSelector(".conditions .info-header i.right");
+	private static final By CONDITIONS_LINK = By.cssSelector(".conditions .info-header i.right");
 
+	private static final By ATTACHMENTS_LINK = By.className("icon-share-alt edit-action right");
+	
 	public ClinicianFacingPatientDashboardPage(Page page) {
 		super(page);
 	}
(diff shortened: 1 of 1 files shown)


sherrif10:
Hello @kdaud  what were you trying to achieve with in this ticket, i thought you were supposed to fix all the changes with in contrib-qa , then if however this is needed, we may need a ticket in distro to handle it

kdaud:
> what were you trying to achieve with in this ticket, i thought you were supposed to fix all the changes with in contrib-qa , then if however this is needed, we may need a ticket in distro to handle it
Oh sory I think i went wrong. 
Here is the use case: The module am working on (contrib-qa) has some  dependency attribute with the distro, so I want to add those functionalities in the distro so that my module in contrib-qa can be enriched . Any advise on this @sherrif10

sherrif10:
You need to fix attachments in contrib qa separately , like how you stated
above, contrib qa is dependent on distro, what you can do is to import the
package you want from distro , or creating that class in contrib qa
specifically

kdaud:
> You need to fix attachments in contrib qa separately , like how you stated above, contrib qa is dependent on distro, what you can do is to import the package you want from distro , or creating that class in contrib qa specifically

Looking at my use case technically, i want to move one of the element (_Attachment_Link_) from attachmentPage in the _contrib-qa_ to _PatientDashboard page_ present in the distro. The better implementation of this element is should be coded in PatientDashboard page rather than in the Attachment Page. By applying this implementation i need to do some technical work with the distro so that i can access this element through a library.
cc: @sherrif10

sherrif10:
You are right with this, However we need to track this change with in separate pull request

sherrif10:
So ideally  to improve code flow and to follow openmrs code flow, you will also create  a new pull request with right branch but following the similar workflow thanks

kdaud:
> So ideally to improve code flow and to follow openmrs code flow, you will also create a new pull request with right branch but following the similar workflow thank

I agree

ui-tests/src/test/java/org/openmrs/reference/page/ClinicianFacingPatientDashboardPage.java
```diff
@@ -453,5 +454,8 @@ public void clickOnAllergiesWidgetLink() {
 	public void clickOnConditionsWidgetLink() {
 		clickOn(CONDITIONS_LINK);
 	}
-
+	
+	public void clickOnAttachmentAppLink() {
+		driver.findElement(ATTACHMENTS_LINK).click();
```
kaweesi, line 459: you can use clickOn rather

kaweesi:
i don't think you need a separate ticket at all, am ok with this same ticket tracking this change in this repo and tracking another change in other repos, that way the same ticket will contain all the changes under development

sherrif10:
@kaweesi  i think you are right  thanks

kdaud:
> am ok with this same ticket tracking this change in this repo and tracking another change in other repos, that way the same ticket will contain all the changes under development

Hey @kaweesi before this information, I had deleted this branch  both from local and on my origin branch and started process on the one that was created to cater for this work. Just as you said, am not able to make changes and update this pr.
However, have recreated one that addressed the previous comment [here](https://github.com/openmrs/openmrs-distro-referenceapplication/pull/304)

k-joseph:
https://github.com/openmrs/openmrs-distro-referenceapplication/pull/304

PR #301, opened 2021-02-28
RATEST-96: Enabled End Visit Test

kdaud:
Here is the ticket id https://issues.openmrs.org/browse/RATEST-96
I have resolved failing tests with EndVistTest


ui-tests/src/test/java/org/openmrs/reference/EndVisitTest.java
```diff
@@ -33,9 +32,8 @@ public void setUp() throws Exception {
     }
 
     @Test
-    @Ignore
     @Category(BuildTests.class)
-    public void EndVisitTest() throws Exception {
+    public void EndVisitTest() {
```
HerbertYiga, line 36: lets track this test again ,dropping the exception shouldn't be the reason why this test was failing

kdaud: > dropping the exception shouldn't be the reason why this test was failing,look more closely on how the method works ie EndVisitTest()

Did you take a look at this [commit](https://github.com/openmrs/openmrs-distro-referenceapplication/pull/301/commits/bdfea9cf519a72bd9d95fc2c7f1007350bc5e4aa).
I think if this [issue](https://issues.openmrs.org/browse/RA-1895) is resolved then this node will be ready

sherrif10:
Compare the[ logs here](https://travis-ci.org/github/openmrs/openmrs-distro-referenceapplication/jobs/760826038) and[ here](https://travis-ci.org/github/openmrs/openmrs-distro-referenceapplication/jobs/760780950) both among your tickets, you will see that almost they are talking about same thing, My advice would be to handle each ticket per ticket otherwise you wont be handle to handle them well successfully and may be fully understanding the cause thanks

kdaud:
> Compare the[ logs here](https://travis-ci.org/github/openmrs/openmrs-distro-referenceapplication/jobs/760826038) and[ here](https://travis-ci.org/github/openmrs/openmrs-distro-referenceapplication/jobs/760780950) both among your tickets, you will see that almost they are talking about same thing, My advice would be to handle each ticket per ticket otherwise you wont be to handle them well successfully and may be fully understanding the cause thanks

Ideally i first began with this issue RATEST-96 but later realized there is something with the host branch. Then opted to first look into it that is why the scenario comes about. Hope you saw this thread on [talk](https://talk.openmrs.org/t/blocker-to-qa-technical-work/32415/5) otherwise I would love to concentrate on one at a time

ui-tests/src/test/java/org/openmrs/reference/EndVisitTest.java
```diff
@@ -33,9 +32,8 @@ public void setUp() throws Exception {
     }
 
     @Test
-    @Ignore
```
kaweesi, line 36: You can run this test locally and fix it, removing this ignore should now be breaking it when you try running it.

kdaud: Taken time to understand this after running the test locally without @ignore! Ideally this test in the EndVistTest was already written But lacks  its definition that why its annotated with @ignore so that it is skipped when running the folder as Junit test. So am required to write its definition, right? This suggests to me that nothing to change in this test except refactoring out the annotation that ignores it after meeting its definition. @k-joseph @HerbertYiga

kaweesi: when a test is ignored, the compiler skips it and runs others not ignored, it's ignored because it should fail when not ignored, work towards fixing it such that it passes when not ignored.

kaweesi: I must also say that keep up the openness, communicating whenever you are blocked or not sure what to do will get you exposed to a lot to learn quickly

kdaud: > when a test is ignored, the compiler skips it and runs others not ignored, it's ignored because it should fail when not ignored, work towards fixing it such that it passes when not ignored.

I get it now! Thanks @k-joseph

kdaud: > I must also say that keep up the openness, communicating whenever you are blocked or not sure what to do will get you exposed to a lot to learn quickly

Welcome!

kaweesi:
Dear @kdaud please update this PR to point to master and fix and pending reviews etc

PR #306, opened 2021-03-03
Initial Testing travis build

sherrif10:
This commit is for testing travis, so there is no much big deal with it , it shall be closed at the end of it thanks everyone

### DIFF
--- a/ui-tests/src/test/java/org/openmrs/reference/AddFormTest.java
+++ b/ui-tests/src/test/java/org/openmrs/reference/AddFormTest.java
@@ -18,7 +18,7 @@ import org.openmrs.uitestframework.test.TestBase;
  */
 public class AddFormTest extends TestBase {
 	
-    private HomePages homePage;
+    private HomePage homePage;
     private HeaderPage headerPage;
     private ManageFormsPage manageForm;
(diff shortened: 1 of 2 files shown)


sherrif10:
cc @k-joseph  am tempted to ping you with a suggestion on this method, so this method was introduced  here https://github.com/openmrs/openmrs-contrib-uitestframework/commit/6fe16add5570b71024a1025f418c839a8f2cbae4, which is bringing the travis to fail, the only reason behind this may be

https://github.com/openmrs/openmrs-contrib-uitestframework/blob/master/src/main/java/org/openmrs/uitestframework/page/Page.java  when on master branch doesnot have this method clickOnLast , and we import org.openmrs.uitestframework.page.Page; when its not checked on qa branch where this method was introduced,

 So i think the quick way to fix this is to write this method in https://github.com/openmrs/openmrs-contrib-uitestframework/blob/master/src/main/java/org/openmrs/uitestframework/page/Page.java when on master branch not on qa branch for us to be able to fetch the method while we import ui.testframework.page in clinician facing patient dashboard. that will fix the error as seem with in this commit 
WIll try to create a pull request following the exact method you had created and we test it out thanks, cant wait to hear from you

PR #309, opened 2021-03-03
RATEST-95:Fix Edit Form Test

HerbertYiga:
Ticket Id:https://issues.openmrs.org/browse/RATEST-95

Description 
Fix Edit Form Test

### DIFF
--- a/ui-tests/src/test/java/org/openmrs/reference/EditFormTest.java
+++ b/ui-tests/src/test/java/org/openmrs/reference/EditFormTest.java
@@ -32,9 +32,8 @@ public class EditFormTest extends TestBase {
         patientDashboardPage = new ClinicianFacingPatientDashboardPage(page);
     }
 
-    @Ignore//ignore due to moving forms functionality
     @Test
-    public void EditFormTest() throws Exception {
+    public void EditFormTest() {
         homePage.goToManageForm();
         if(!manageForm.addPresent()) {
             manageForm.delete();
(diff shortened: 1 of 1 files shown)


sherrif10:
Same as here, it would automatically pass because it is considered as ignored, so try un ingoring it and push again thanks

ui-tests/src/test/java/org/openmrs/reference/EditFormTest.java
```diff
@@ -32,9 +32,8 @@ public void setUp() throws Exception {
         patientDashboardPage = new ClinicianFacingPatientDashboardPage(page);
     }
 
-    @Ignore//ignore due to moving forms functionality
     @Test
-    public void EditFormTest() throws Exception {
+    public void EditFormTest() {
```
ellieMayVelasquez, line 36: I detect that this code is problematic. According to the [Bad practice (BAD_PRACTICE)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#bad-practice-bad-practice), [Nm: Method names should start with a lower case letter (NM_METHOD_NAMING_CONVENTION)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#nm-method-names-should-start-with-a-lower-case-letter-nm-method-naming-convention).
Methods should be verbs, in mixed case with the first letter lowercase, with the first letter of each internal word capitalized.

HerbertYiga: i will have to close this and make a better pr for this ticket

ellieMayVelasquez: Noticed :)

PR #312, opened 2021-03-03
RATEST-86:Fix Eye Report Test

HerbertYiga:
Ticket ID:https://issues.openmrs.org/browse/RATEST-86

Description
Fix Eye Report Test


sherrif10:
Thanks @HerbertYiga  does this fail when you remove exception, did you clearly see the @ignore annotation , We need to get rid of @ignore annotations for these tests to run successfully thanks , try removing it and push again thanks
i think cross check all the commits  that you have pushed currently with reference to this suggestion thanks,

PR #325, opened 2021-03-10
Fix Typo Error On Administration Page 

sherrif10:
cc @k-joseph  fixed typo error , This has also been fixed on legacy ui here https://issues.openmrs.org/browse/LUI-176


sherrif10:
Hi @k-joseph  i think this should be  re opened because it seems you didnt address it please check it here  https://github.com/openmrs/openmrs-distro-referenceapplication/blob/master/ui-tests/src/test/java/org/openmrs/reference/page/AdministrationPage.java  thanks

PR #326, opened 2021-03-10
RATEST-74:Failing Test Ignored AddFormTests

sherrif10:
Wow @k-joseph  thanks alot for the work

### DIFF
--- a/ui-tests/src/test/java/org/openmrs/reference/AddFormTest.java
+++ b/ui-tests/src/test/java/org/openmrs/reference/AddFormTest.java
@@ -37,7 +37,7 @@ public class AddFormTest extends  ReferenceApplicationTestBase {
     public void addFormTest() throws Exception {
     	administrationPage.goToManageForms();
     	assertPage(administrationPage);
-        if(!manageForm.addPresent()) {
+        if (!manageForm.addPresent()) {
 
             manageForm.delete();
         }
--- a/ui-tests/src/test/java/org/openmrs/reference/AddFormTest.java
+++ b/ui-tests/src/test/java/org/openmrs/reference/AddFormTest.java
@@ -6,7 +6,6 @@ import org.junit.After;
 import org.junit.Before;
 import org.openmrs.reference.groups.BuildTests;
 import org.junit.experimental.categories.Category;
-//import org.junit.Ignore;
 import org.junit.Test;
 import org.junit.experimental.categories.Category;
 import org.openmrs.reference.groups.BuildTests;
@@ -34,11 +33,9 @@ public class AddFormTest extends  ReferenceApplicationTestBase {
          manageForm = new ManageFormsPage(driver);
     }
 
-//    @Ignore //ignore due to moving forms functionality
     @Test
     @Category(BuildTests.class)
     public void addFormTest() throws Exception {
-//        homePage.goToManageForm();
     	administrationPage.goToManageForms();
     	assertPage(administrationPage);
         if(!manageForm.addPresent()) {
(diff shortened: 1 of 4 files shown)


ui-tests/src/test/java/org/openmrs/reference/AddFormTest.java
```diff
@@ -4,37 +4,40 @@
...
+
 
     @Before
     public void setUp() throws Exception {
-        homePage = new HomePage(page);
-        assertPage(homePage);
-        headerPage = new HeaderPage(driver);
-        manageForm = new ManageFormsPage(driver);
-        patientDashboardPage = new ClinicianFacingPatientDashboardPage(page);
+    	 administrationPage = new AdministrationPage(page);
```
HerbertYiga, line 30: thanks @sharif i find starting from the home page makes traversing the tests orderly ie step by step, we could actually start with ( homePage = new HomePage(page);

sherrif10: Why would you start with homePage

HerbertYiga: > Why would you start with homePage, i started with AdministrationPage because am considering adminPage as the parent page of manageForms

well starting from the home page makes the flow of tests orderly than jumping to any other page since we are simulating the user experience

sherrif10: Thanks @HerbertYiga  , i dont think that every page must start from homepage to be simulated , i also think we can adopt inheritance approach which has been used from child parents, eg manageFormPage having child attributes like editForm, deleteForm,saveForm , so ideally thats the same approach that we can adopt  cc @k-joseph  any suggestion on this

sherrif10: @HerbertYiga , May be to me more clear Taking a look here [homePage](https://github.com/openmrs/openmrs-distro-referenceapplication/blob/master/ui-tests/src/test/java/org/openmrs/reference/page/HomePage.java) attributes, there is a link to manage Forms which we need  not to consider because that page is different from manageForms on[ administrationPage,](https://github.com/openmrs/openmrs-distro-referenceapplication/blob/master/ui-tests/src/test/java/org/openmrs/reference/page/AdministrationPage.java), dont know if i make sense

kaweesi:
dear @sherrif10 update this to resolve merge issues

ui-tests/src/test/java/org/openmrs/reference/AddFormTest.java
kaweesi: as long as the test workflow isn't broken, finding the right page and testing the right functionality is all that matters. If you can ensure the prior steps to getting to the page are reused, that would be still ok

ui-tests/src/test/java/org/openmrs/reference/page/AdministrationPage.java
```diff
@@ -11,17 +11,19 @@
...
 import org.openmrs.uitestframework.page.Page;
 import org.openqa.selenium.By;
+import org.openqa.selenium.WebDriver;
 
 public class AdministrationPage extends Page {
 
     private static final String URL = "/admin/index.htm";
 
-    private static final By MANAGE_USERS = By.cssSelector("#legacyui-mangeUsers a");
+    private static final By MANAGE_USERS = By.cssSelector("#legacyui-manageUsers a");
```
kaweesi, line 20: please move this to its own corresponding ticket and attach commits to legacy repo to it.

sherrif10: ok thanks let me create its own separate ticket

ui-tests/src/test/java/org/openmrs/reference/AddFormTest.java
sherrif10: sure thats what was trying to achieve

PR #337, opened 2021-03-18
WIP:RATEST-74:Failing Test Ignored AddFormTests

sherrif10:
cc @k-joseph  , This is WIP

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ addons:
   sauce_connect:
     username: $SAUCELABS_USERNAME
     access_key: $SAUCELABS_ACCESSKEY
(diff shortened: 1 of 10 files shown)


sherrif10:
Hello @k-joseph  is there anyone who could be getting similar error, This seems to be like not from my code

kdaud (commented):
Thanks for this effort @sherrif10 .Do you mind changing the property name of webdriver from _webdriver.gecko.driver_ to _webdriver.chrome.driver_  present in test.properties file since we are redirecting to chrome driver not firefox. Lets watch what will happen with travis again

sherrif10:
Sure will update that use case

PR #361, opened 2021-04-25
RATEST-120: Fix Add Provider Test

kdaud:
Ticket ID: https://issues.openmrs.org/browse/RATEST-120
Description
Fix Add Provider Test in legacy ui

### DIFF
--- a/ui-tests/src/test/java/org/openmrs/reference/AddProviderTest.java
+++ b/ui-tests/src/test/java/org/openmrs/reference/AddProviderTest.java
@@ -11,7 +11,7 @@ package org.openmrs.reference;
 
 import org.junit.After;
 import org.junit.Before;
-import org.junit.Ignore;
+import org.junit.Test;
 import org.junit.experimental.categories.Category;
 import org.openmrs.reference.groups.BuildTests;
 import org.openmrs.reference.page.AdministrationPage;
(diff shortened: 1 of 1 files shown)


ui-tests/src/test/java/org/openmrs/reference/AddProviderTest.java
```diff
@@ -26,14 +26,14 @@
...
-    public void setUp() {
+	public void setUp() throws Exception {
         person = TestData.generateRandomPerson();
         personUuid = TestData.createPerson(this.person);
     }
 
-    @Ignore
+	@Test
     @Category(BuildTests.class)
-    public void addProviderTest() throws InterruptedException {
```
sherrif10, line 36: LGTM

sherrif10 (commented):
cc @kaweesi

ui-tests/src/test/java/org/openmrs/reference/AddProviderTest.java
sherrif10, line 36: However this  doesnot change anything

sherrif10 (approved):

ui-tests/src/test/java/org/openmrs/reference/AddProviderTest.java
kdaud: > However this doesnot change anything

According to method implementation, there other exceptions besides InterruptException that can occur and so  using the BaseClass _Exception_ would be a better approach rather than the child class

ui-tests/src/test/java/org/openmrs/reference/AddProviderTest.java
```diff
@@ -26,14 +26,14 @@
     private String personUuid;
 
     @Before
-    public void setUp() {
+	public void setUp() throws Exception {
```
kaweesi, line 29: clean up the formatting issues, same comment here, any need to throwing the exception here?

PR #373, opened 2021-05-03
RATEST-124: Retrieving the user Selected-Location element

kdaud:
Ticket ID: https://issues.openmrs.org/browse/RATEST-124

Description
Failure when retrieving the user Selected-Location element in distro during test builds


sherrif10:
Had same idea , However was skeptical, i think we can as well increase the  runnable thread in Page class  to 3000L 
  Like

     **try {
            Thread.sleep(3000L);
            this.waiter.until(expectation);
        } catch (Throwable var3) {
            Assert.fail("Timeout waiting for Page.");
        }**
However  this LGTM   thanks @kdaud  cc @k-joseph

PR #380, opened 2021-05-06
RATEST-131: Fix Delete Allergy Test

kdaud:
Ticket ID https://issues.openmrs.org/browse/RATEST-131

Description:
Fix Delete Allergy Test

### DIFF
--- a/ui-tests/src/test/java/org/openmrs/reference/DeleteAllergyTest.java
+++ b/ui-tests/src/test/java/org/openmrs/reference/DeleteAllergyTest.java
@@ -12,7 +12,7 @@ package org.openmrs.reference;
 
 import org.junit.After;
 import org.junit.Before;
-import org.junit.Ignore;
+import org.junit.Test;
 import org.junit.experimental.categories.Category;
 import org.openmrs.reference.groups.BuildTests;
 import org.openmrs.reference.page.ActiveVisitsPage;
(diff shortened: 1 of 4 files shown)


ui-tests/src/test/java/org/openmrs/reference/DeleteAllergyTest.java
```diff
@@ -31,12 +31,12 @@
     @Before
     public void setUp() throws Exception {
         patient = createTestPatient();
-        createTestVisit();
+        new TestData.TestVisit(patient.uuid, TestData.getAVisitType(), getLocationUuid(homePage)).create();
```
sherrif10, line 34: Thanks @kdaud  , why are we adding this createTestVisit methd in the before part

kdaud: > why are we adding this createTestVisit methd in the before part , what happens if we dont touch that part

As far as I know **we definitely need this method**. Its not a good to expose out test to any of the data-set that could be present within the server. So this method ensures data-set creation and the test work flow for which the test will consume rather than using any external data. cc: @sherrif10

ui-tests/src/test/java/org/openmrs/reference/page/ClinicianFacingPatientDashboardPage.java
```diff
@@ -288,11 +287,6 @@ public String findPatientId() {
         return findElement(PATIENT_ID).getText();
     }
 
-    public AllergyPage clickOnAllergyManagement() {
```
kdaud, line 291: For software quality and readability, Have replaced the implementation of clickOnAllergyManagement()  with clickOnAllergiesWidgetLink() since they all perform the same goal. Updated the respective child classes within the module that implement the functionality of the page. However, am skeptical about other modules elsewhere that use this functionality through a library But for the sake of software quality its a good option to make some refactoring in our base module

k-joseph:
Please update this PR with master and rerun the build

PR #383, opened 2021-05-09
RATEST-132: Fix Capture Vitals Test

kdaud:
Ticket ID: https://issues.openmrs.org/browse/RATEST-132

Description:
Fix Capture Vitals Test

### DIFF
--- a/ui-tests/src/test/java/org/openmrs/reference/CaptureVitalsTest.java
+++ b/ui-tests/src/test/java/org/openmrs/reference/CaptureVitalsTest.java
@@ -11,7 +11,7 @@ package org.openmrs.reference;
 
 import org.junit.After;
 import org.junit.Before;
-import org.junit.Ignore;
+import org.junit.Test;
 import org.junit.experimental.categories.Category;
 import org.openmrs.reference.groups.BuildTests;
 import org.openmrs.reference.page.ActiveVisitsPage;
@@ -35,9 +35,9 @@ public class CaptureVitalsTest extends LocationSensitiveApplicationTestBase {
         deletePatient(patient);
     }
 
-    @Ignore
+    @Test
     @Category(BuildTests.class)
-    public void captureVital() throws InterruptedException {
+    public void captureVitalsTest() {
 
         ActiveVisitsPage activeVisitsPage = homePage.goToActiveVisitsSearch();
         activeVisitsPage.search(patient.identifier);
--- a/ui-tests/src/test/java/org/openmrs/reference/page/ClinicianFacingPatientDashboardPage.java
+++ b/ui-tests/src/test/java/org/openmrs/reference/page/ClinicianFacingPatientDashboardPage.java
@@ -31,7 +31,7 @@ public class ClinicianFacingPatientDashboardPage extends Page {
     private static final By CONFIRM = By.cssSelector("#quick-visit-creation-dialog .confirm");
     private static final By STARTED_AT = By.className("active-visit-started-at-message");
     private static final By VISIT_NOTE = By.linkText("Visit Note");
-    private static final By PATIENT_CAPTURE_VITALS = By.id("referenceapplication.realTime.vitals");
+    private static final By PATIENT_CAPTURE_VITALS = By.linkText("Capture Vitals");
     private static final By DIAGNOSIS_SEARCH_CONTAINER = By.id("diagnosis-search-container");
     private static final By DIAGNOSIS_SEARCH = By.id("diagnosis-search");
     private static final By VISIT_LINK_2 = By.className("visit-link");


ui-tests/src/test/java/org/openmrs/reference/CaptureVitalsTest.java
```diff
@@ -27,17 +27,12 @@
...
-    @After
-    public void tearDown() throws Exception {
-        deletePatient(patient);
+        new TestData.TestVisit(patient.uuid, TestData.getAVisitType(), getLocationUuid(homePage)).create();
     }
 
-    @Ignore
+    @Test
     @Category(BuildTests.class)
-    public void captureVital() throws InterruptedException {
```
sherrif10, line 40: is this a typo error, does it break when it remains captureVital

kdaud: > is this a typo error, does it break when it remains captureVital

No!. However its our role as QA Engineer to align to the OpenMRS software standards(conventions). Actually its supposed to be _captureVitalsTest()_

sherrif10: Thats the exact answer i wanted from you , thanks and well done

ui-tests/src/test/java/org/openmrs/reference/page/ClinicianFacingPatientDashboardPage.java
```diff
@@ -31,7 +31,7 @@
     private static final By CONFIRM = By.cssSelector("#quick-visit-creation-dialog .confirm");
     private static final By STARTED_AT = By.className("active-visit-started-at-message");
     private static final By VISIT_NOTE = By.id("referenceapplication.realTime.simpleVisitNote");
```
kaweesi, line 33: this should also be affected

ui-tests/src/test/java/org/openmrs/reference/page/ClinicianFacingPatientDashboardPage.java
```diff
@@ -31,7 +31,7 @@
...
-    private static final By PATIENT_CAPTURE_VITALS = By.id("referenceapplication.realTime.vitals");
+    private static final By PATIENT_CAPTURE_VITALS = By.linkText("Capture Vitals");
```
kaweesi, line 34: replace with an element selector
`By.cssSelector("a[id='referenceapplication.realTime.vitals']")`

ui-tests/src/test/java/org/openmrs/reference/page/ClinicianFacingPatientDashboardPage.java
kdaud: > this should also be affected

_Boolean value=true_

ui-tests/src/test/java/org/openmrs/reference/page/ClinicianFacingPatientDashboardPage.java
kdaud: Thanks @kaweesi . Caught this in the updated commit

kdaud: After simulating the test, the element PATIENT_CAPTURE_VITALS is retrievable by the existing attribute.

kdaud: After simulating the test, the element PATIENT_CAPTURE_VITALS is retrievable by the existing attribute.

k-joseph: what did u finally do with this?

kdaud: The element is retrievable by the existing set attribute. Changing to _cssSelector_ could be alternative to retrieving the element. @k-joseph should I go for _cssSelector_

kdaud: This is @kdaud  once again! 
After simulating the behaviour of the element (PATIENT_CAPTURE_VITALS) a couple of times, I observed that the element fails to be retrieved by selenium API with the existing attribute. Took time to use different look ups via is (i_d,cssSelector, xpath, and others_) and realised the element only fails with chrome headless =true. Simulated a use case when retrieving the element via By.cssSelector("a[id='referenceapplication.realTime.vitals']")  and similar failure occurs as well. The best look up for this element is via _linkText()_ which passes on both chrome and firefox instances. cc: @k-joseph

PR #388, opened 2021-05-10
RATEST-113: Failing Test on Firefox

kdaud:
Ticket ID: https://issues.openmrs.org/browse/RATEST-113

Description:
Tests classes that consumes ManageProviderScheduleTest fails on firefox engine

### DIFF
--- a/ui-tests/src/test/java/org/openmrs/reference/ManageProviderSchedulesTest.java
+++ b/ui-tests/src/test/java/org/openmrs/reference/ManageProviderSchedulesTest.java
@@ -39,7 +39,7 @@ public class ManageProviderSchedulesTest extends ReferenceApplicationTestBase {
 	@Before
 	public void setUp() throws Exception {
 		homePage = new HomePage(page);
-		assertPage(homePage);
+		assertPage(homePage.waitForPage());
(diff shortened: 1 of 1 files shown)


ui-tests/src/test/java/org/openmrs/reference/ManageProviderSchedulesTest.java
```diff
@@ -39,7 +39,6 @@
 	@Before
 	public void setUp() throws Exception {
 		homePage = new HomePage(page);
-		assertPage(homePage);
```
kdaud, line 42: Since the homePage is already instantiated by the object _homePage = new HomePage(page)_ then there is no need to assertPage(homePage)

ui-tests/src/test/java/org/openmrs/reference/ManageProviderSchedulesTest.java
```diff
@@ -39,7 +39,7 @@
...
+		assertPage(homePage.waitForPage());
```
kdaud, line 42: For `assertPage()`to be effective in checking the current page, we need to wait the `homePage` to load

PR #392, opened 2021-05-12
RATEST-138: Fix Delete Visit Type Test

kdaud:
Ticket ID: https://issues.openmrs.org/browse/RATEST-138

Description:
Failing Test on firefox: DeleteVisitTypeTest

### DIFF
--- a/ui-tests/src/test/java/org/openmrs/reference/page/VisitTypeListPage.java
+++ b/ui-tests/src/test/java/org/openmrs/reference/page/VisitTypeListPage.java
@@ -45,6 +45,7 @@ public class VisitTypeListPage extends AdminManagementPage {
 
     public List<String> getVisitTypeList() {
         List<String> visitTypeList = new ArrayList<String>();
+        driver.navigate().refresh();
         for (WebElement webElement : findElements(VISIT_TYPE_LIST)) {
             visitTypeList.add(webElement.getText());
         }
--- a/ui-tests/src/test/java/org/openmrs/reference/page/VisitTypeListPage.java
+++ b/ui-tests/src/test/java/org/openmrs/reference/page/VisitTypeListPage.java
@@ -12,6 +12,7 @@ package org.openmrs.reference.page;
 import org.openmrs.uitestframework.page.Page;
 import org.openqa.selenium.By;
 import org.openqa.selenium.WebElement;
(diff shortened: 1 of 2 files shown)


ui-tests/src/test/java/org/openmrs/reference/page/VisitTypeListPage.java
```diff
@@ -45,6 +45,7 @@ public VisitTypePage goToVisitType(String name) {
 
     public List<String> getVisitTypeList() {
         List<String> visitTypeList = new ArrayList<String>();
+        driver.navigate().refresh();
```
kaweesi, line 48: any reason why u want to refresh the page here?

kdaud: The element goes Into stale condition; either the element is read s if its no longer attached to the DOM, or Its read as if it is not in the current frame context. Refreshing the page makes the element accessible

k-joseph: Are u only able to resolve that this way only?

kdaud: A smart way for the resolution is in the incoming push.

ui-tests/src/test/java/org/openmrs/reference/page/VisitTypeListPage.java
```diff
@@ -45,7 +46,7 @@ public VisitTypePage goToVisitType(String name) {
 
     public List<String> getVisitTypeList() {
         List<String> visitTypeList = new ArrayList<String>();
-        driver.navigate().refresh();
```
kdaud: This does the required magic and sorts out the element to be retrievable

ui-tests/src/test/java/org/openmrs/reference/page/VisitTypeListPage.java
```diff
@@ -45,7 +46,7 @@ public VisitTypePage goToVisitType(String name) {
...
+        waiter.until(ExpectedConditions.visibilityOfElementLocated(VISIT_TYPE_LIST));
```
kdaud, line 49: This is smart enough to make the element retrievable

PR #393, opened 2021-05-13
RATEST-140: Fix Find Patient Record Test

kdaud:
Ticket ID: https://issues.openmrs.org/browse/RATEST-140

Description:
Fix Find Patient Record Test


kdaud:
Messed up with referencing to the right PR for better github workflow. Deleting the reference from the parent comment does not make the child node loose the root

PR #402, opened 2021-05-24
RATEST-147: ci format workflow fails on forks

kdaud:
Ticket ID: https://issues.openmrs.org/browse/RATEST-147

Description
ci format workflow fails on forks. See the logs [here](https://github.com/kdaud/openmrs-contrib-uitestframework/runs/2653112821?check_suite_focus=true#step:2:6)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,7 +2,7 @@ name: Legacy Selenium Firefox
 on:
   push:
     branches: [master]
-  pull_request:
+  pull_request_target:
     branches: [master]
   workflow_dispatch:
 jobs:


sherrif10:
Thanks for working on this @kdaud ,  Did you recognise that you disabled  firefox githubactions

kdaud:
Have not recognized it actually. @sherrif10 do you mind pointing me the _code line_ to this. Thanks

.github/workflows/ci.yml
```diff
@@ -2,7 +2,7 @@ name: Legacy Selenium Firefox
 on:
   push:
     branches: [master]
-  pull_request:
+  pull_request_target:
```
sherrif10, line 5: Probably reverting your change might make it run again, isnt this fix supposed to be on uitestframework not on distro

kdaud: > isnt this fix supposed to be on uitestframework not on distro

Sure! @sherrif10 would you recommend shifting these changes to _contrib_ module to make this resolved

sherrif10: You would probably need to create another pull request in contrib uitestframework i think , however looking forward to hear from cc @k-joseph  about this work around

kdaud: This event pull_request_target was introduced by GitHub recently in order to support the PR from forked repositories. Here is the [documentation ](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull_request_target).

kaweesi:
Have you been affected by this already? none should be blocked by this, the build should run ok i think. BTW, have u tried opening a PR against this change and confirmed that pull_request_target behaves the same way as pull_request?

kdaud:
> Have you been affected by this already?

I noticed as my commits are passing with distro, I get noticed through an email for the build failure on the uitestframework module for the PR that shows green on distro

kdaud:
> none should be blocked by this, the build should run ok i think.

Boolean value=true

kdaud:
> have u tried opening a PR against this change and confirmed that pull_request_target behaves the same way as pull_request?

Am simulating this on one of my personal modules and see the if there exists a difference on the two. Will feedback after observation

kaweesi:
BTW, one reason why i wouldn't wish to allow this for PRs is that until the PR is merged no need to inform depending on applications which is what that workflow does. You should only be facing issues if u try pushing changes directly to your master branch of your fork

kdaud:
> until the PR is merged no need to inform depending on applications which is what that workflow does

I get the logic! I agree

kaweesi:
We will close this as not needed for now

PR #412, opened 2021-06-02
RATEST-135: Fix Add User Test

kdaud:
Ticket ID: https://issues.openmrs.org/browse/RATEST-135

Description
Fix add user test

### DIFF
--- a/ui-tests/src/test/java/org/openmrs/reference/page/AdministrationPage.java
+++ b/ui-tests/src/test/java/org/openmrs/reference/page/AdministrationPage.java
@@ -16,7 +16,7 @@ public class AdministrationPage extends Page {
 
     private static final String URL = "/admin/index.htm";
 
-    private static final By MANAGE_USERS = By.cssSelector("#legacyui-manageUsers a");
+    private static final By MANAGE_USERS = By.cssSelector("#content a[href='/openmrs/admin/users/users.list']");
     private static final By MANAGE_MODULES = By.cssSelector("#legacyui-manageModules a");
     private final static By MANAGE_VISIT_TYPES = By.cssSelector("#legacyui-manageVisitTypes a");
     private static final By MANAGE_PROVIDERS = By.cssSelector("#content a[href='/openmrs/admin/provider/index.htm']");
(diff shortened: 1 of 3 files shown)


ui-tests/src/test/java/org/openmrs/reference/page/AdministrationPage.java
```diff
@@ -16,7 +16,7 @@
 
     private static final String URL = "/admin/index.htm";
 
-    private static final By MANAGE_USERS = By.cssSelector("#legacyui-manageUsers a");
```
sherrif10, line 19: Any reason why changing to the defined full url, i think this should be left as a tag not defined content url

kdaud: > Any reason why changing to the defined full url

Investigation report indicates clearly that the current property attribute for the [element](https://github.com/openmrs/openmrs-distro-referenceapplication/blob/285dcc90731099bd2d9368097428a21b3c8bcca2/ui-tests/src/test/java/org/openmrs/reference/page/AdministrationPage.java#L19) does not exist. Apart from the suggested property, the element can also be retrieved via its `id, xpath, linkText and others` but in terms of` elementLoadTime`, the suggested one takes less time to retrieve the element than any other attribute can do. Putting quality assurance in mind, the suggested property is given a priority

sherrif10: > Investigation Lab report indicates clearly that the current property attribute for the [element](https://github.com/openmrs/openmrs-distro-referenceapplication/blob/285dcc90731099bd2d9368097428a21b3c8bcca2/ui-tests/src/test/java/org/openmrs/reference/page/AdministrationPage.java#L19) does not exist.

The element does not exist ??

kdaud: > The element does not exist ??

Lab report indicates clearly that the current property attribute for the [element](https://github.com/openmrs/openmrs-distro-referenceapplication/blob/285dcc90731099bd2d9368097428a21b3c8bcca2/ui-tests/src/test/java/org/openmrs/reference/page/AdministrationPage.java#L19)

PR #421, opened 2021-06-02
Failing Test on firefox: LoginTest

sherrif10:
https://issues.openmrs.org/browse/RATEST-111

### DIFF
--- a/ui-tests/src/test/java/org/openmrs/reference/LoginTest.java
+++ b/ui-tests/src/test/java/org/openmrs/reference/LoginTest.java
@@ -16,67 +16,66 @@ import org.openmrs.reference.page.HomePage;
 
 import static org.junit.Assert.assertTrue;
 
-public class LoginTest extends ReferenceApplicationTestBase {
-    
+public class LoginTest extends ReferenceApplicationTestBase { 
     private HomePage homePage;
 
     public void initiateHomePage() {
-	homePage = new HomePage(page);
-	assertPage(homePage.waitForPage());
+        homePage = new HomePage(page);
+        assertPage(homePage.waitForPage());
     }
 
     @Test
     @Category(BuildTests.class)
     public void verifyModulesAvailableOnHomePage() {
-	goToLoginPage().loginAsAdmin();
-	initiateHomePage();
-	assertTrue(homePage.isFindAPatientAppPresent());
-	assertTrue(homePage.isActiveVisitsAppPresent());
-	assertTrue(homePage.isAppointmentSchedulingAppPresent());
-	assertTrue(homePage.isCaptureVitalsAppPresent());
-	assertTrue(homePage.isRegisterPatientCustomizedForRefAppPresent());
-	assertTrue(homePage.isDataManagementAppPresent());
-	assertTrue(homePage.isConfigureMetadataAppPresent());
-	assertTrue(homePage.isSystemAdministrationAppPresent());
+        goToLoginPage().loginAsAdmin();
+        initiateHomePage();
+        assertTrue(homePage.isFindAPatientAppPresent());
+        assertTrue(homePage.isActiveVisitsAppPresent());
+        assertTrue(homePage.isAppointmentSchedulingAppPresent());
+        assertTrue(homePage.isCaptureVitalsAppPresent());
+        assertTrue(homePage.isRegisterPatientCustomizedForRefAppPresent());
+        assertTrue(homePage.isDataManagementAppPresent());
+        assertTrue(homePage.isConfigureMetadataAppPresent());
+        assertTrue(homePage.isSystemAdministrationAppPresent());
     }
 
     @Test
     @Category(BuildTests.class)
     public void verifyClerkModulesAvailableOnHomePage() {
-	goToLoginPage().loginAsClerk();
-	initiateHomePage();
-	assertTrue(homePage.isActiveVisitsAppPresent());
-	assertTrue(homePage.isAppointmentSchedulingAppPresent());
-	assertTrue(homePage.isRegisterPatientCustomizedForRefAppPresent());
+        goToLoginPage().loginAsClerk();
+        initiateHomePage();
+        assertTrue(homePage.isActiveVisitsAppPresent());
+        assertTrue(homePage.isAppointmentSchedulingAppPresent());
+        assertTrue(homePage.isRegisterPatientCustomizedForRefAppPresent());
     }
 
     @Test
     @Category(BuildTests.class)
     public void verifyDoctorModulesAvailableOnHomePage() {
-	goToLoginPage().loginAsDoctor();
-	initiateHomePage();
-	assertTrue(homePage.isFindAPatientAppPresent());
-	assertTrue(homePage.isActiveVisitsAppPresent());
-	assertTrue(homePage.isAppointmentSchedulingAppPresent());
+        goToLoginPage().loginAsDoctor();
+        initiateHomePage();
+        assertTrue(homePage.isFindAPatientAppPresent());
+        assertTrue(homePage.isActiveVisitsAppPresent());
+        assertTrue(homePage.isAppointmentSchedulingAppPresent());
     }
 
     @Test
     @Category(BuildTests.class)
     public void verifyNurseModulesAvailableOnHomePage() {
-	goToLoginPage().loginAsNurse();
-	initiateHomePage();
-	assertTrue(homePage.isFindAPatientAppPresent());
-	assertTrue(homePage.isActiveVisitsAppPresent());
-	assertTrue(homePage.isAppointmentSchedulingAppPresent());
-	assertTrue(homePage.isCaptureVitalsAppPresent());
+        goToLoginPage().loginAsNurse();
+        initiateHomePage();
+        assertTrue(homePage.isFindAPatientAppPresent());
+        assertTrue(homePage.isActiveVisitsAppPresent());
+        assertTrue(homePage.isAppointmentSchedulingAppPresent());
+        assertTrue(homePage.isCaptureVitalsAppPresent());
     }
 
     @Test
     @Category(BuildTests.class)
     public void verifySysadminModulesAvailableOnHomePage() {
-	goToLoginPage().loginAsSysadmin();
-	initiateHomePage();
-	assertTrue(homePage.isAppointmentSchedulingAppPresent());
-	assertTrue(homePage.isSystemAdministrationAppPresent());
-    }  
+        goToLoginPage().loginAsSysadmin();
+        initiateHomePage();
+        assertTrue(homePage.isAppointmentSchedulingAppPresent());
+        assertTrue(homePage.isSystemAdministrationAppPresent());
+    }
 }
--- a/ui-tests/src/test/java/org/openmrs/reference/LoginTest.java
+++ b/ui-tests/src/test/java/org/openmrs/reference/LoginTest.java
@@ -17,6 +17,7 @@ import org.openmrs.reference.page.HomePage;
 import static org.junit.Assert.assertTrue;
 
 public class LoginTest extends ReferenceApplicationTestBase {
+    
     private HomePage homePage;
 
     public void initiateHomePage() {
@@ -47,7 +48,6 @@ public class LoginTest extends ReferenceApplicationTestBase {
 	assertTrue(homePage.isActiveVisitsAppPresent());
 	assertTrue(homePage.isAppointmentSchedulingAppPresent());
 	assertTrue(homePage.isRegisterPatientCustomizedForRefAppPresent());
-
     }
 
     @Test
@@ -78,6 +78,5 @@ public class LoginTest extends ReferenceApplicationTestBase {
 	initiateHomePage();
 	assertTrue(homePage.isAppointmentSchedulingAppPresent());
 	assertTrue(homePage.isSystemAdministrationAppPresent());
-    }
-
+    }  
 }
--- a/ui-tests/src/test/java/org/openmrs/reference/LoginTest.java
+++ b/ui-tests/src/test/java/org/openmrs/reference/LoginTest.java
@@ -9,70 +9,75 @@
  */
 package org.openmrs.reference;
 
-import org.junit.Ignore;
 import org.junit.Test;
 import org.junit.experimental.categories.Category;
+import org.openmrs.reference.groups.BuildTests;
 import org.openmrs.reference.page.HomePage;
 
 import static org.junit.Assert.assertTrue;
 
-@Ignore
 public class LoginTest extends ReferenceApplicationTestBase {
     private HomePage homePage;
 
-    @Test
-    @Category(org.openmrs.reference.groups.BuildTests.class)
-    public void verifyModulesAvailableOnHomePage() throws Exception {
-        homePage = new HomePage(page);
-        assertPage(homePage);
-        assertTrue(homePage.isFindAPatientAppPresent());
-        assertTrue(homePage.isActiveVisitsAppPresent());
-        assertTrue(homePage.isAppointmentSchedulingAppPresent());
-        assertTrue(homePage.isCaptureVitalsAppPresent());
-        assertTrue(homePage.isRegisterPatientCustomizedForRefAppPresent());
-        assertTrue(homePage.isDataManagementAppPresent());
-        assertTrue(homePage.isConfigureMetadataAppPresent());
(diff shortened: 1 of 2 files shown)


ui-tests/src/test/java/org/openmrs/reference/LoginTest.java
```diff
@@ -12,19 +12,20 @@
...
     private HomePage homePage;
-
+    
     @Test
-    @Category(org.openmrs.reference.groups.BuildTests.class)
+    @Category(BuildTests.class)
     public void verifyModulesAvailableOnHomePage() throws Exception {
+    	goToLoginPage().loginAsAdmin();
         homePage = new HomePage(page);
-        assertPage(homePage);
```
kdaud, line 27: any reasons for removing these _asserPage()_

kdaud: The function is very vital  as it checks the current page see https://github.com/openmrs/openmrs-contrib-uitestframework/blob/master/src/main/java/org/openmrs/uitestframework/test/TestBase.java#L340

sherrif10: Sure it checks the current Page using that method but the current page here already by default has been reached , initialising homePage with in the class creates a new instance which is by default homePage, its like a repetition of asserting page(homePage)

kdaud: > the current page here already by default has been reached

What would happen if the page does not exist and the function assertPage() is no longer present

sherrif10: Didnot get your point well but ideally  homePage is inherited from loginPage, the login workflow is basing on homePage, whatever we are testing falls within homepage, we cant say after logging in the automation is redirected to back unless its a different test

kdaud: > Didnot get your point well

Take a `use case` when the page does not exist and the function `assertPage()` is no longer present as suggested in the your PR. How will this be handled ?

sherrif10: The page will by default exist ofcourse like i said above these pages are paginating to each other, when you say page doesnot exist , it means loginfunction wil not be working

kdaud: As removing `assertPage()` may seem to be the solution, But this will cause a lot of issues in the long run. So I recommend to retain this method within the class.

sherrif10: Before you recommend please could you throw tell us which issues do you think it might cause in the long run thanks

kdaud: > please could you tell us which issues do you think it might cause

Refer https://github.com/openmrs/openmrs-contrib-uitestframework/blob/master/src/main/java/org/openmrs/uitestframework/test/TestBase.java#L340

sherrif10: ok thank you let me look through that , but thats not the problem i guess , suppose you are asserting the page when you are already on the expected page

kdaud: The `function` is very vital as it checks the current page and it should not be omitted.

sherrif10: ok thanks for your review

kdaud: work around ensuring the test passes with the presence of this function rather than getting rid of it.

kdaud: > but thats not the problem i gues
Add `waitForPage()`  function on `homePage` object to ensure the page is loaded for assertion to be effective. Something like `assertPage(homePage..waitForPage());`

sherrif10: You have still not getting my point, even though we add a wait element tag, the assertion expection  wont change

kdaud: > even though we add a wait element tag, the assertion expection wont change

Have you taken a use case on this

sherrif10: Havent tried waiting for homePage still wondering why waiting for homePage when its throwing assertion error migh change its assertion to pass

kdaud: > Havent tried waiting for homePage

Does it still throw the exception?

sherrif10: its working thanks though practically it does not add up

kdaud: > though practically it does not add up

Refer to my recent response [here](https://issues.openmrs.org/browse/RATEST-93)

ui-tests/src/test/java/org/openmrs/reference/LoginTest.java
```diff
@@ -25,6 +25,7 @@
     public void verifyModulesAvailableOnHomePage() throws Exception {
```
kdaud, line 25: Remove the `throws`

ui-tests/src/test/java/org/openmrs/reference/LoginTest.java
```diff
@@ -39,6 +40,7 @@ public void verifyModulesAvailableOnHomePage() throws Exception {
     public void verifyClerkModulesAvailableOnHomePage() throws Exception {
```
kdaud, line 40: Even here

ui-tests/src/test/java/org/openmrs/reference/LoginTest.java
```diff
@@ -48,6 +50,7 @@ public void verifyClerkModulesAvailableOnHomePage() throws Exception {
     public void verifyDoctorModulesAvailableOnHomePage() throws Exception {
```
kdaud, line 50: And `here`

ui-tests/src/test/java/org/openmrs/reference/LoginTest.java
```diff
@@ -57,6 +60,7 @@ public void verifyDoctorModulesAvailableOnHomePage() throws Exception {
     public void verifyNurseModulesAvailableOnHomePage() throws Exception {
```
kdaud, line 60: Here as well

ui-tests/src/test/java/org/openmrs/reference/LoginTest.java
```diff
@@ -67,6 +71,7 @@ public void verifyNurseModulesAvailableOnHomePage() throws Exception {
     public void verifySysadminModulesAvailableOnHomePage() throws Exception {
```
kdaud, line 71: And here` also`

sherrif10: True i aggree since adding a waitForPage handles that

kdaud:
Add   `@Category(BuildTests.class)` annotation before `verifyClerkModulesAvailableOnHomePage()` , `verifyDoctorModulesAvailableOnHomePage()`, `verifyNurseModulesAvailableOnHomePage()`, verifySysadminModulesAvailableOnHomePage()

ui-tests/src/test/java/org/openmrs/reference/LoginTest.java
```diff
@@ -36,41 +37,41 @@ public void verifyModulesAvailableOnHomePage() throws Exception {
...
     @Test
-    public void verifyClerkModulesAvailableOnHomePage() throws Exception {
+    public void verifyClerkModulesAvailableOnHomePage() {
         goToLoginPage().loginAsClerk();
         homePage = new HomePage(page);
-        assertPage(homePage);
+        assertPage(homePage.waitForPage());
         assertTrue(homePage.isActiveVisitsAppPresent());
         assertTrue(homePage.isAppointmentSchedulingAppPresent());
-        assertTrue(homePage.isRegisterPatientCustomizedForRefAppPresent());
```
kdaud, line 45: Any reason for removing this

ui-tests/src/test/java/org/openmrs/reference/LoginTest.java
```diff
@@ -36,41 +37,41 @@ public void verifyModulesAvailableOnHomePage() throws Exception {
...
 
     @Test
-    public void verifyClerkModulesAvailableOnHomePage() throws Exception {
+    public void verifyClerkModulesAvailableOnHomePage() {
         goToLoginPage().loginAsClerk();
         homePage = new HomePage(page);
-        assertPage(homePage);
+        assertPage(homePage.waitForPage());
         assertTrue(homePage.isActiveVisitsAppPresent());
         assertTrue(homePage.isAppointmentSchedulingAppPresent());
```
kdaud, line 45: Any reason for removing this

ui-tests/src/test/java/org/openmrs/reference/LoginTest.java
kdaud, line 45: Any reason for removing this

ui-tests/src/test/java/org/openmrs/reference/LoginTest.java
```diff
@@ -13,10 +13,8 @@
 import org.junit.experimental.categories.Category;
 import org.openmrs.reference.groups.BuildTests;
 import org.openmrs.reference.page.HomePage;
-
```
kdaud, line 16: This space is needed for code quality and readability

sherrif10: The space here is not neccessary ,  i guess

kdaud:
Since the codeBlock below is consumed by many functions, for software quality assurance I recommend you refactor it out  
`{
homePage = new HomePage(page)
assertPage(homePage.waitforPage())
}`
from all the  functions that consumes it and create one function

`public void initiateHomePage()
{
homePage = new HomePage(page)
assertPage(homePage.waitforPage())
}

and then invoke the function `initiateHomePage()` into those methods that consumes it.

sherrif10:
Thanks for the suggestion but that might need another ticket since it will need to refactor the entire tests that are using this kind of method.

ui-tests/src/test/java/org/openmrs/reference/LoginTest.java
kdaud: According to our community import conventions, `devs` are encouraged to separate different categories of imports by a one space line between them. Look through other existing classes on how imports are organized

sherrif10: Uhmm whats the difference between these  two 
**import org.openmrs.reference.groups.BuildTests;
import org.openmrs.reference.page.HomePage;**

public void loginTest(){}
And 
import org.openmrs.reference.groups.BuildTests;

**import org.openmrs.reference.page.HomePage;
public void loginTest(){
}** 
which of these looks smarter

kaweesi:
good suggestion but i don't think all those places require the current page being homePage.

sherrif10:
Thats definitely true, @kdaud  you might have your response on this

sherrif10:
cc @k-joseph  @kdaud  is there anything to address with in this pull request or more comments to resolve

kdaud:
> but i don't think all those places require the current page being homePage.

Within the same class, the above code block have shared is consumed by five functions;  `verifyModulesAvailableOnHomePage(), verifyClerkModulesAvailableOnHomePage(), verifyDoctorModulesAvailableOnHomePage(), verifyNursesModulesAvailableOnHomePage(), verifySysModulesAvailableOnHomePage() `. Its from this point that I made the recommendation above to ensure we have  codebase that aligns with software quality assurance

sherrif10:
cc @kaweesi  , any response on @kdaud  suggestion

kaweesi:
@sherrif10 then @kdaud 's suggestion is valid to avoid duplication

ui-tests/src/test/java/org/openmrs/reference/LoginTest.java
kdaud: Have not got you well. But my reference was at line number 16

sherrif10: ok thanks

sherrif10:
Let me address the review above thanks

ui-tests/src/test/java/org/openmrs/reference/LoginTest.java
```diff
@@ -9,70 +9,75 @@
...
-        assertTrue(homePage.isAppointmentSchedulingAppPresent());
-        assertTrue(homePage.isRegisterPatientCustomizedForRefAppPresent());
+    @Category(BuildTests.class)
+    public void verifyClerkModulesAvailableOnHomePage() {
+	goToLoginPage().loginAsClerk();
+	initiateHomePage();
+	assertTrue(homePage.isActiveVisitsAppPresent());
+	assertTrue(homePage.isAppointmentSchedulingAppPresent());
+	assertTrue(homePage.isRegisterPatientCustomizedForRefAppPresent());
+
```
kdaud, line 50: Remove this redundant line `Line Number 50`

ui-tests/src/test/java/org/openmrs/reference/LoginTest.java
```diff
@@ -9,70 +9,75 @@
...
-        assertPage(homePage);
-        assertTrue(homePage.isAppointmentSchedulingAppPresent());
-        assertTrue(homePage.isSystemAdministrationAppPresent());
+    @Category(BuildTests.class)
+    public void verifySysadminModulesAvailableOnHomePage() {
+	goToLoginPage().loginAsSysadmin();
+	initiateHomePage();
+	assertTrue(homePage.isAppointmentSchedulingAppPresent());
+	assertTrue(homePage.isSystemAdministrationAppPresent());
     }
```
kdaud, line 82: Remove this as well for code quality `(Line Number 82)`

kdaud:
Thanks @sherrif10 for the great work. Catch my last reviews and this will be ready to enter into master.

ui-tests/src/test/java/org/openmrs/reference/LoginTest.java
sherrif10: Why is it agood Practice to add a line at the end of the file (EOF), i think this is the  question that we should be asking ourselves , Practically , Eclipse, intellij IDE  default adds a trailing line at the EOF this improves code readability by other tools like git especially using diff function in git, FWIW  this has been motivated by core developers in openmrs , [feel free to checkout why we need a line at the EOF](https://stackoverflow.com/questions/4700312/empty-new-line-at-the-end-of-the-java-source-files)

kdaud: > Why is it agood Practice to add a line at the end of the file (EOF

IMO for code readability and its a good practice in software development

>  this improves code readability by other tools like git especially using diff function in git

I strongly agree

> FWIW this has been motivated by core developers in openmrs so we need to keep that as well

I agree with this convention.  And this trailing last line after the closer of  the last file curl bracket as you pointed it clearly (EOF) should be maintained.

However,  I was referring to the redundant line located at `line Number 82 `  NOT  `Line Number 84`






`

sherrif10:
Any more comments cc @kdaud  @kaweesi

kdaud:
I think your changes satisfies the ticket requirement. 

However, you seem to have mixed up things with your commits for example did you intend to include the commit with id  `84cd1e7`  (the current third last commit: `RATEST-145: Fix End Visit Test`) I can see this [commit](https://github.com/openmrs/openmrs-distro-referenceapplication/pull/421/commits/84cd1e7136b1e84450184f1f40ed94e412475c95) is not yet in master

I would recommend to clean up the PR to only show  all your changes

sherrif10:
Opps merge conflicts brought that , thanks for the catch

sherrif10:
@kdaud  do you still have more suggestions or comments cc @kaweesi

kdaud:
> do you still have more suggestions or comments

Other than catching my previous [review](https://github.com/openmrs/openmrs-distro-referenceapplication/pull/421#issuecomment-861156338), I stand to say thanks  for the great work @sherrif10

sherrif10:
Thanks too

PR #436, opened 2021-06-13
RATEST-141: Fix find patient by name test

kdaud:
Ticket ID: https://issues.openmrs.org/browse/RATEST-141

Description
Fix find patient by name test


kdaud:
`Waiting the page to load` ensures that the entered patient identifier matches with the test created `patient identifier`. Otherwise without `waiting the page to load`, the comparison is made against the `searched patient identifier `with any of the existing patient in the `db server instance `instead of the` test created patient`

PR #454, opened 2021-06-17
RATEST-135: Fix add user test

kdaud:
Ticket [ID](https://issues.openmrs.org/browse/RATEST-135)
Description -> Fix add user test

### DIFF
--- a/ui-tests/src/test/java/org/openmrs/reference/AddUserTest.java
+++ b/ui-tests/src/test/java/org/openmrs/reference/AddUserTest.java
@@ -15,7 +15,11 @@ import java.util.List;
 import org.junit.Test;
 import org.junit.experimental.categories.Category;
 import org.openmrs.reference.groups.BuildTests;
-import org.openmrs.reference.page.*;
+import org.openmrs.reference.page.AddEditUserPage;
+import org.openmrs.reference.page.AdministrationPage;
+import org.openmrs.reference.page.HomePage;
+import org.openmrs.reference.page.ManageUserPage;
+import org.openmrs.reference.page.SystemAdministrationPage;
 
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertTrue;
--- a/ui-tests/src/test/java/org/openmrs/reference/AddUserTest.java
+++ b/ui-tests/src/test/java/org/openmrs/reference/AddUserTest.java
@@ -10,12 +10,12 @@
 
 package org.openmrs.reference;
 
+import java.util.List;
+
 import org.junit.Test;
 import org.junit.experimental.categories.Category;
 import org.openmrs.reference.groups.BuildTests;
 import org.openmrs.reference.page.*;
-
-import java.util.List;
 
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertTrue;
--- a/ui-tests/src/test/java/org/openmrs/reference/AddUserTest.java
+++ b/ui-tests/src/test/java/org/openmrs/reference/AddUserTest.java
@@ -10,7 +10,9 @@
 
 package org.openmrs.reference;
 
-import org.junit.Ignore;
+import org.junit.Test;
+import org.junit.experimental.categories.Category;
+import org.openmrs.reference.groups.BuildTests;
 import org.openmrs.reference.page.*;
 
 import java.util.List;
@@ -20,8 +22,9 @@ import static org.junit.Assert.assertTrue;
 
 public class AddUserTest extends ReferenceApplicationTestBase {
 
-    @Ignore
-    public void addUserTest() throws Exception {
+    @Test
+    @Category(BuildTests.class)
+    public void addUserTest() {
         SystemAdministrationPage systemAdministrationPage = homePage.goToSystemAdministrationPage();
         AdministrationPage administrationPage = systemAdministrationPage.goToAdvancedAdministration();
         ManageUserPage manageUserPage = administrationPage.clickOnManageUsers();
@@ -55,6 +58,7 @@ public class AddUserTest extends ReferenceApplicationTestBase {
             assertTrue(manageUserPage.getUserSavedNotification().contains("User Saved"));
 
             homePage = new HomePage(goToLoginPage().login("super_nurse", "Nurse123"));
+            homePage.waitForPage();
             homePage.getLoggedUsername().contains("super_nurse");
             goToLoginPage().loginAsAdmin();
             homePage.goToAdministration();
--- a/ui-tests/src/test/java/org/openmrs/reference/page/AdministrationPage.java
+++ b/ui-tests/src/test/java/org/openmrs/reference/page/AdministrationPage.java
@@ -15,7 +15,7 @@ import org.openqa.selenium.By;
 public class AdministrationPage extends Page {
 
     private static final String URL = "/admin/index.htm";
-    private static final By MANAGE_USERS = By.cssSelector("#legacyui-manageUsers a");
+    private static final By MANAGE_USERS = By.cssSelector("#content a[href='/openmrs/admin/users/users.list']");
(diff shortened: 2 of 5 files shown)


ui-tests/src/test/java/org/openmrs/reference/AddUserTest.java
```diff
@@ -55,6 +58,7 @@ public void addUserTest() throws Exception {
             assertTrue(manageUserPage.getUserSavedNotification().contains("User Saved"));
 
             homePage = new HomePage(goToLoginPage().login("super_nurse", "Nurse123"));
+            homePage.waitForPage();
```
kdaud, line 61: Though the text to be checked by the function  `getLoggedUsername()` is loaded in `DOM ` after the login operation But it takes longer to be rendered on the `UI` which makes it to be invisible during the action `homePage.getLoggedUsername().contains("super_nurse"); `

ui-tests/src/test/java/org/openmrs/reference/page/AdministrationPage.java
```diff
@@ -16,7 +16,7 @@
 
     private static final String URL = "/admin/index.htm";
 
-    private static final By MANAGE_USERS = By.cssSelector("#legacyui-manageUsers a");
+    private static final By MANAGE_USERS = By.cssSelector("#content a[href='/openmrs/admin/users/users.list']");
```
kdaud, line 19: Inspection Lab report indicates that the current property attribute for the [element](https://github.com/openmrs/openmrs-distro-referenceapplication/blob/285dcc90731099bd2d9368097428a21b3c8bcca2/ui-tests/src/test/java/org/openmrs/reference/page/AdministrationPage.java#L19) does not exist. Apart from the suggested property, the element can also be retrieved via its `id, xpath, linkText and others` but in terms of` elementLoadLoadTime`, the suggested one takes less time to retrieve the element than any other attribute can do thus overriding the rest.

ui-tests/src/test/java/org/openmrs/reference/page/HomePage.java
```diff
@@ -43,7 +44,7 @@ private boolean isAppButtonPresent(String appId) {
     }
 
     public String getLoggedUsername() {
-        return findElement(By.cssSelector(".identifier")).getText();
```
kdaud, line 46: Inspection on the element shows there are more than one element using the current value property. The suggested attribute property uniquely identifies the element thus overriding

ui-tests/src/test/java/org/openmrs/reference/page/HomePage.java
```diff
@@ -43,7 +44,7 @@ private boolean isAppButtonPresent(String appId) {
...
+        return findElement(LOGGED_IN_USER).getText();
```
kdaud, line 47: Inspection on the `element` in context shows there are more than one element using the current value property. The suggested `attribute value property` uniquely identifies the element thus overriding.

ui-tests/src/test/java/org/openmrs/reference/page/AdministrationPage.java
```diff
@@ -16,7 +16,7 @@
 
     private static final String URL = "/admin/index.htm";
 
-    private static final By MANAGE_USERS = By.cssSelector("#legacyui-manageUsers a");
```
sherrif10, line 19: Since this element have its id and we can as well use its ccselector, i think  we  can still maintain  **"#legacyui-manageUsers a** for code redability , since the full url  can easily change but id wont  change

kdaud: Thanks for the feedback !! Have wrapped the response in the commit with id `1107113`

ui-tests/src/test/java/org/openmrs/reference/AddUserTest.java
kdaud, line 61: Though the text to be checked by the function `getLoggedUsername() `is loaded in `DOM` after the login operation But it takes longer to be rendered on the `UI` which makes it to be invisible during the action `homePage.getLoggedUsername().contains("super_nurse");`.

ui-tests/src/test/java/org/openmrs/reference/page/HomePage.java
kdaud, line 47: Inspection on the `element` in context shows there are more than one element using the current `value property`. The suggested` attribute value property` uniquely identifies the element thus overriding.

ui-tests/src/test/java/org/openmrs/reference/page/HomePage.java
```diff
@@ -29,6 +29,7 @@
     private static final By FIND_PATIENT_RECORD = By.id("coreapps-activeVisitsHomepageLink-coreapps-activeVisitsHomepageLink-extension");
     private static final By DATA_MANAGEMENT = By.id("coreapps-datamanagement-homepageLink-coreapps-datamanagement-homepageLink-extension");
     private static final By APPOINTMENT_SCHEDULING = By.id("appointmentschedulingui-homeAppLink-appointmentschedulingui-homeAppLink-extension");
+    private static final By LOGGED_IN_USER = By.xpath("//*[@id='navbarSupportedContent']/ul/li[1]");
```
k-joseph, line 32: if i am to accept an xpath, for this, i would start from the ul at-least or else use a better reference that may hardly break

ui-tests/src/test/java/org/openmrs/reference/AddUserTest.java
```diff
@@ -10,18 +10,21 @@
...
 
-import org.junit.Ignore;
-import org.openmrs.reference.page.*;
-
 import java.util.List;
 
+import org.junit.Test;
+import org.junit.experimental.categories.Category;
+import org.openmrs.reference.groups.BuildTests;
+import org.openmrs.reference.page.*;
```
k-joseph, line 18: explicitly import classes since this would avail all pages which is not right

ui-tests/src/test/java/org/openmrs/reference/page/HomePage.java
kdaud: > you may need to start from .identifier and check it's children to target not the icon but the label

I opted for a reliable `attribute property value` that may not easily break to ensure code sustainability and wrapped it into a commit with id `2421397`

ui-tests/src/test/java/org/openmrs/reference/page/HomePage.java
kdaud: This has been addressed [here](https://github.com/openmrs/openmrs-distro-referenceapplication/pull/454#discussion_r662772257)

ui-tests/src/test/java/org/openmrs/reference/AddUserTest.java
kdaud: > explicitly import classes since this would avail all pages which is not right

This is absolutely true!! Have addressed this in the commit with id` 2421397`

ui-tests/src/test/java/org/openmrs/reference/page/AdministrationPage.java
kdaud: > i think we can still maintain **"#legacyui-manageUsers a** for code redability

This would be the best option among the alternatives. However, this element attribute value is only true when inspected via qa instance, and inspecting it via `demo.openmrs.org` & also on the` sdk local instance` the element property value is not true. In those two instances have mentioned the element `value property` is set to `#legacyui-mangeUsers a`. For code sustainability and quality assurance, I opted to set the property value to the suggested one since it hardly breaks with any server instance either



> full url can easily change

This ideally might not be true!!

kdaud:
The commit with id `120d45d` might seem to be out of the scope But felt important to refactor the code base in those two java class that were overlooked within in this [ticket](https://issues.openmrs.org/browse/RATEST-160) for quality assurance

PR #457, opened 2021-06-18
RATEST-160: Refactoring legacy ui-tests code base for quality assurance

kdaud:
Ticket [ID](https://issues.openmrs.org/browse/RATEST-160)
Description:
Refactor the legacy ui tests in respect to the following:
- Removing unwanted `throws`
- Removing redundant `spaces` within the codebase for quality assurance
- Removing unused `imports`
- Organizing `imports` to align with the organization software standards
- Aligning test method `naming` convention

See details [here](https://wiki.openmrs.org/display/docs/Java+Conventions)

### DIFF
--- a/ui-tests/src/test/java/org/openmrs/reference/AddVisitTypeTest.java
+++ b/ui-tests/src/test/java/org/openmrs/reference/AddVisitTypeTest.java
@@ -21,8 +21,6 @@ import org.openmrs.reference.page.VisitTypePage;
 import static org.hamcrest.CoreMatchers.hasItem;
 import static org.hamcrest.MatcherAssert.assertThat;
 
-/**
- */
 public class AddVisitTypeTest extends ReferenceApplicationTestBase {
 
     private String visitTypeName;
@@ -36,7 +34,7 @@ public class AddVisitTypeTest extends ReferenceApplicationTestBase {
 
     @Test
     @Category(BuildTests.class)
-    public void addVisitTypeTest() throws InterruptedException {
+    public void addVisitTypeTest() {
         AdministrationPage administrationPage = homePage.goToAdministration();
         VisitTypeListPage visitTypeListPage = administrationPage.goToVisitTypePage();
         VisitTypePage visitTypePage = visitTypeListPage.addVisitType();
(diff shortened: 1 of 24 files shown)


ui-tests/src/test/java/org/openmrs/reference/AddVisitTypeTest.java
```diff
@@ -36,7 +34,7 @@ public void setUp() {
 
     @Test
     @Category(BuildTests.class)
-    public void addVisitTypeTest() throws InterruptedException {
```
sherrif10, line 39: This is good, however i think we need to be keen in the ways we handle exceptions, i dont  think that we need to remove this exceptions  in each line, From my experience interrupted exceptions matters alot  especially handling threading,when a thread /element is waiting , i would recommend to remove exception would come along specific ticket

kdaud: > i think we need to be keen in the ways we handle exceptions, i don't think that we need to remove this exceptions in each line

Good suggestion. AFAICT exceptions are very vital to be handled at method level and shouldn't be neglected at All if its expected. 


> i would recommend to remove exception would come along specific ticket

AFAIK there are some `throws` that we do not need at all on this test methods. However, there are some which we definitely need to handle exceptions that may be encountered when invoking functions which within them exception can occur. Just as you mentioned that `InterruptedException` is vey important which `I didn't know before`. I agree that test methods with such throws can keep hold of them and  those that may not need them (Exception) we get rid of the `throws. `

 As QA is concerned, I felt important that the old legacy ui tests codebase is refactored to align with the software standards and also improve our organization software as I see in other modules. This compelled me to dig into the code base to carry out the technical work as stressed in the description above and wrapped the work in the PR. 

If we can identify the `specific` throws that I may need to revert, I am willing to dedicate time for this But the code base `ui-tests` need a clean up action to align with software standards in reference to quality assurance. 

I would love to hear form @sherrif10 @k-joseph about your opinion on the epic

k-joseph:
these refactoring are a good idea to me, thanks @kdaud for thinking about these

sherrif10:
well done @kdaud

PR #461, opened 2021-06-19
RATEST-143: Fix record merge issue test

kdaud:
Ticket[ ID](https://issues.openmrs.org/browse/RATEST-143)
Description -> Fix record merge issue test

### DIFF
--- a/ui-tests/src/test/java/org/openmrs/reference/RecordMergeIssueTest.java
+++ b/ui-tests/src/test/java/org/openmrs/reference/RecordMergeIssueTest.java
@@ -12,12 +12,12 @@ import org.openmrs.uitestframework.test.TestData;
 
 import static org.junit.Assert.assertFalse;
 
-
 /**
  * Created by tomasz on 09.07.15.
  */
 
 public class RecordMergeIssueTest extends TestBase {
+
     private HomePage homePage;
     private HeaderPage headerPage;
     private FindPatientPage findPatientPage;
@@ -28,7 +28,6 @@ public class RecordMergeIssueTest extends TestBase {
     private DataManagementPage dataManagementPage;
     private String id;
     private String id2;
-
 
     @Before
     public void setUp() throws Exception {
@@ -74,7 +73,6 @@ public class RecordMergeIssueTest extends TestBase {
         dataManagementPage.clickOnContinue();
         assertFalse(driver.getPageSource().contains("java.lang.NullPointerException"));
     }
-
 
     @After
     public void tearDown() throws Exception {
--- a/ui-tests/src/test/java/org/openmrs/reference/RecordMergeIssueTest.java
+++ b/ui-tests/src/test/java/org/openmrs/reference/RecordMergeIssueTest.java
@@ -2,7 +2,9 @@ package org.openmrs.reference;
 
 import org.junit.After;
 import org.junit.Before;
-import org.junit.Ignore;
+import org.junit.Test;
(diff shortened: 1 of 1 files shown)


ui-tests/src/test/java/org/openmrs/reference/RecordMergeIssueTest.java
```diff
@@ -80,6 +83,5 @@ public void tearDown() throws Exception {
         p.uuid = patient.uuid;
         deletePatient(p);
         waitForPatientDeletion(patient.uuid);
-        headerPage.logOut();
```
kdaud, line 83: After deleting the test patient data from the `db instance` there is no need to invoke `headerPage.logOut()`  within `tearDown() ` function

ui-tests/src/test/java/org/openmrs/reference/RecordMergeIssueTest.java
```diff
@@ -41,9 +43,10 @@ public void setUp() throws Exception {
         patient1 = new TestPatient();
     }
 
-    @Ignore
+    @Test
+    @Category(BuildTests.class)
     public void recordMergeIssueTest() throws Exception {
-        homePage.goToRegisterPatientApp();
+        homePage.goToRegisterPatientApp().waitForPage();
```
kdaud, line 48: After the test invoking many pages during the test `setUp()`, the patient registration  page takes a bit longer to be loaded  by the `classLoader`. Adding `waitforPage()` resolves the delay **Otherwise** the actions that follow are not effective and the patient data is not saved in `db instance`

ui-tests/src/test/java/org/openmrs/reference/RecordMergeIssueTest.java
sherrif10: Why not Adding header.logout, do you mean after deletion of patient, and the test wait to delete a patient  the testpasses without logging out ??

kdaud: > Why not Adding header.logout

This being a `teardown` function, I think there is no need to invoke `headerPage.logout()`. Basing on the legacy ui test `workflow`, the test comes to an end when the `test patient data` has been deleted from `db instance` by the logic of `RestClient`. By the time   `headerPage.logOut()` is invoked, the` test patient data` has been already deleted from the `db instance` at  LN 84 & 85


> do you mean after deletion of patient, and the test wait to delete a patient the testpasses without logging out ??

After simulation on the test without invoking  `headerPage.logOut()` within the `teardown` function, the test `passes`.

ui-tests/src/test/java/org/openmrs/reference/RecordMergeIssueTest.java
```diff
@@ -41,9 +42,10 @@ public void setUp() throws Exception {
         patient1 = new TestPatient();
     }
 
-    @Ignore
+    @Test
+    @Category(BuildTests.class)
     public void recordMergeIssueTest() throws Exception {
-        homePage.goToRegisterPatientApp();
```
sherrif10, line 46: is this change necessary,

kdaud: > is this change necessary,

Explanation about the change is already [here](https://github.com/openmrs/openmrs-distro-referenceapplication/pull/461#discussion_r654835119)

PR #467, opened 2021-06-21
Bundling SPA Module

HerbertYiga:
Bundling the spa module cc @dkayiwa


dkayiwa:
How about here? https://github.com/openmrs/openmrs-distro-referenceapplication/blob/master/package/src/main/resources/openmrs-distro.properties

PR #474, opened 2021-06-25
Reworking of 3.x configuration

ibacher:
This is the version of the distro I've been working on. It's not the cleanest thing ever, but it should both support building images via CI as well as the current workflow.

I've changed quite a few things, obviously. Some changes for compatibility with how things are currently structured (e.g., using org.openmrs.distro rather than org.openmrs as the groupId; versions are shared across the components, etc.). Other things are to reflect evolving naming conventions (in the SDK, we opted, after some discussion, to rename the microfrontend stuff to "spa"). I've also reworked things so that the spa parts are served from /openmrs/spa, as they are on the current server. Note, though, that the spa is still served via the UI Apache server.

Finally, I've adapted the [docker-openmrs-server](https://github.com/PIH/docker-openmrs-server) image that @mseaton put together. This image uses the same distribution structure that this project generates, but it's quite a bit more customisable than the default OMRS image.

Future work:

* Ideally, there would be a single "skinny" image which could be loaded with a distribution the way the docker-compose file currently does things and the "fat" embedded image for OMRS's infrastructure would be built on top of that. Currently, though, they are two different images that just use the same files.
* It would be good to have some compose examples for, e.g., Postgres, MariaDB, and MySQL. Currently, this just runs on MariaDB.
* The frontend doesn't quite work right now, but this mostly seems to be doen to missing metadata. The nice part about starting from something as simple as this image is we can start to get really clear about what metadata needs to be there for the SPA environment to work.

@rbuisson @mseaton If you have any comments, I'd really appreciate them, though I realise this is probably too big for a detailed review.


package/pom.xml
```diff
@@ -61,22 +43,21 @@
     <appframeworkVersion>2.16.0</appframeworkVersion>
 
     <!-- OpenMRS Core -->
-    <openMRSVersion>2.3.3</openMRSVersion>
-
+    <openMRSVersion>2.4.0</openMRSVersion>
```
rbuisson, line 46: I am suprised that this works. Last time I tried Web Services REST didn't work with 2.4.0 🤔

README.md
```diff
@@ -12,7 +12,7 @@ cd run/docker
 docker-compose up
 ```
 
-OpenMRS Microfrontends UI is accessible at http://localhost/ui
+New OpenMRS UI is accessible at http://localhost/openmrs/spa
```
rbuisson, line 15: I am suprised that the call doesn't get intercepted by the Apache ProxyPass directive for `/openmrs`...

run/docker/proxy/proxy.conf
```diff
@@ -1,15 +1,17 @@
 <VirtualHost *:80>
+  # Frontend
```
rbuisson, line 2: I see you've moved this higher up so that it catches the `/openmrs/spa` before `/openmrs` . Got it now!

package/spa-config.json
```diff
@@ -1,28 +1,31 @@
...
-    "@openmrs/esm-patient-test-results-app": "next"
+    "@openmrs/esm-patient-notes-app": "next",
+    "@openmrs/esm-patient-programs-app": "next",
+    "@openmrs/esm-patient-registration-app": "next",
+    "@openmrs/esm-patient-search-app": "next",
+    "@openmrs/esm-patient-test-results-app": "next",
+    "@openmrs/esm-patient-vitals-app": "next",
+    "@openmrs/esm-primary-navigation-app": "next"
+  },
+  "not-working-mfs": {
```
rbuisson, line 27: Is this doing something specific? or just a place to stack ESMs we don't use.

package/pom.xml
```diff
@@ -61,22 +43,21 @@
...
   </properties>
 
   <dependencies>
     <dependency>
-      <groupId>org.openmrs</groupId>
+      <groupId>${project.groupId}</groupId>
       <artifactId>openmrs-config-refapp</artifactId>
       <type>zip</type>
-      <version>${openmrsConfigVersion}</version>
+      <version>${project.version}</version>
```
rbuisson, line 54: The reason I didn't use the same versions for Maven modules is that I thought we may ultimately want to remove the configs from this project and have them to live on their own repo.

That way if one just wants to change a single piece of metadata, only the OpenMRS config will have to released. SPA, SPA configs and other artifacts won't need to be re-released/re-deployed.

But I guess it doesn't make sense if they are all in the same repo.

package/pom.xml
```diff
@@ -248,41 +238,47 @@
...
-            <phase>initialize</phase>
+            <phase>compile</phase>
             <configuration>
-              <executable>npm</executable>
+              <executable>npx</executable>
               <arguments>
-                <argument>install</argument>
                 <argument>openmrs@${appShellVersion}</argument>
+                <argument>build</argument>
+                <argument>--spa-path</argument>
```
rbuisson, line 273: I guess that will solve [the issue Burke and others](https://talk.openmrs.org/t/creating-a-build-pipeline-for-openmrs-3-0/33719/12?u=mksrom) had with environment vars being substituted at build time.

rbuisson (commented):
Thanks @ibacher , this looks great!

> The frontend doesn't quite work right now, but this mostly seems to be due to missing metadata

Yes, there is currently [this PR](https://github.com/openmrs/openmrs-distro-referenceapplication/pull/401) on the way. Hopefully that'll fix most of the issues.
And there is also the `next` tags often don't work.

package/spa-config.json
ibacher: It's just a place to store them. It's intended to be temporary and mostly as a reminder to myself to bug the MF guys about them.

package/pom.xml
ibacher: Yeah... to be honest, I was a little unsure about this, because there could be real value in versioning the metadata separately from the plumbing and certainly for something other than the RefApp.

package/pom.xml
```diff
@@ -207,34 +185,46 @@
...
       </plugin>
 
       <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-antrun-plugin</artifactId>
-        <version>1.7</version>
         <executions>
           <execution>
-            <id>rename-jars-to-omod</id>
+            <id>rename jars to omod</id>
```
mseaton, line 227: We shouldn't need to rename jars to omod, or rename the war file.  We should be using the sdk plugin based on the openmrs-distro.properties file to pull down the war, modules, and owas that are defined and rename them appropriately.

mseaton (commented):
A few comments...

package/README.md
```diff
@@ -11,28 +11,32 @@ Output:
...
 #### OpenMRS modules
 `omod`s are specified as Maven `<dependency>` in the [pom.xml](pom.xml) file.
 
-#### OpenMRs configuration (Initializer)
+#### OpenMRS Configuration (Initializer)
 OpenMRS config can be set in [openmrs-config/configuration/](openmrs-config/configuration/) folder
 
 #### Micro Frontends
-ESMs are set in the [microfrontends.json](microfrontends.json) file.
+ESMs are set in the [spa-config.json](spa-config.json) file.
```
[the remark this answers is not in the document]
ibacher: Yep... that's being addressed in a future PR against this branch. (This is from before that feature was released for the SDK, considering I just finished releasing it now).

package/README.md
```diff
@@ -11,28 +11,32 @@ Output:
...
 
 ----
 
 ### Specifying dependencies
 #### OpenMRS modules
 `omod`s are specified as Maven `<dependency>` in the [pom.xml](pom.xml) file.
 
-#### OpenMRs configuration (Initializer)
+#### OpenMRS Configuration (Initializer)
 OpenMRS config can be set in [openmrs-config/configuration/](openmrs-config/configuration/) folder
```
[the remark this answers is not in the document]
ibacher: `openmrs-config` and `spa-config`, at least in this iteration, are actually Maven sub-projects. But, yeah, eventually that would be possible.

package/README.md
```diff
@@ -11,28 +11,32 @@ Output:
...
+  The Configuration for OpenMRS frontend.
 * `openmrs_modules/`
-<br/>The required set of OpenMRS modules.
+  
+  The required set of OpenMRS modules.
 * `openmrs_config/`
-<br/>The OpenMRS bespoke configuration (more [here](https://github.com/mekomsolutions/openmrs-config-haiti)) to be processed by the [Initializer module](https://github.com/mekomsolutions/openmrs-module-initializer).
+  
+  The OpenMRS local configuration (more [here](https://github.com/mekomsolutions/openmrs-config-haiti)) to be processed by the [Initializer module](https://github.com/mekomsolutions/openmrs-module-initializer).
 * `openmrs_core/`
```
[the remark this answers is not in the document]
ibacher: So, I could see the advantage of having additional webapps deployed alongside OMRS in a non-Dockerised environment. But, it seems to me that having multiple webapps running inside a Dockerised Tomcat container is something of an anti-pattern we wouldn't want to encourage.

package/README.md
```diff
@@ -11,28 +11,32 @@ Output:
...
+#### OpenMRS Configuration (Initializer)
 OpenMRS config can be set in [openmrs-config/configuration/](openmrs-config/configuration/) folder
 
 #### Micro Frontends
-ESMs are set in the [microfrontends.json](microfrontends.json) file.
+ESMs are set in the [spa-config.json](spa-config.json) file.
 
 #### Micro Frontends configuration
-MF Config can be set in [microfrontends-config/configuration/](microfrontends-config/configuration/) folder
+MF Config can be set in [spa-config/configuration/](spa-config/configuration/) folder
```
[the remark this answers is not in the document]
ibacher: So, the MF / SPA project does have a concept of a deployment-specific configuration that includes things like the exact deployment-specific layout of MFs or MF-specific configurations. Although I don't know that this current version handles that appropriately, we *do* want to be able to deploy that in a actual location.

package/README.md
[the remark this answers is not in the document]
ibacher: Which isn't to say that this isn't how things work... anything in the `openmrs_core` folder still gets deployed to the Tomcat webapps folder... I just think its not behaviour we want to encourage.

package/README.md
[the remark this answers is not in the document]
brandones: I'm having trouble parsing this, sorry.

MF has two kinds of "config"—the build config and the application config. I assume this is for application config. Build config (`microfrontends.json` then `spa-config.json`) is now integrated into distro.properties. The build process needs to provide a URL to `openmrs build` at which the application config files will be served. So I assume that the stuff in `spa-config/configuration` eventually gets served.

PR #479, opened 2021-06-30
RATEST-93:Failing Test: Delete Form Test

sherrif10:
https://issues.openmrs.org/browse/RATEST-93  cc @k-joseph  Kindly review thanks

### DIFF
--- a/ui-tests/src/test/java/org/openmrs/reference/DeleteFormTest.java
+++ b/ui-tests/src/test/java/org/openmrs/reference/DeleteFormTest.java
@@ -2,15 +2,15 @@ package org.openmrs.reference;
 
 import org.junit.After;
 import org.junit.Before;
-import org.junit.Ignore;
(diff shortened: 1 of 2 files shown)


sherrif10:
This should fail again on Ci  untill AddForm Test get merged simply because AddForm Test adds the form by default, so if the ci runs this test, it will throw run Time exceptions because there is no form to delete in the instance

k-joseph:
> This should fail again on Ci if its merged first , until Add Form Test get Fixed probably merged simply because Add Form Test adds the form by default, so if the ci runs this test, it will throw run Time exceptions hence posing somewhere where the element of delete Form doesnot exist only because there is no form to delete in the instance

Please make this test completely independent

sherrif10:
Hello @k-joseph  breaking other tests on the ci i think we might need to revert it again

PR #482, opened 2021-07-01
RATEST-111:Failing Test on firefox: LoginTest

sherrif10:
https://issues.openmrs.org/browse/RATEST-111  cc @k-joseph  @kdaud

### DIFF
--- a/ui-tests/src/test/java/org/openmrs/reference/LoginTest.java
+++ b/ui-tests/src/test/java/org/openmrs/reference/LoginTest.java
@@ -9,22 +9,26 @@
  */
 package org.openmrs.reference;
 
-import org.junit.Ignore;
 import org.junit.Test;
 import org.junit.experimental.categories.Category;
+import org.openmrs.reference.groups.BuildTests;
 import org.openmrs.reference.page.HomePage;
 
 import static org.junit.Assert.assertTrue;
 
-@Ignore
+
 public class LoginTest extends ReferenceApplicationTestBase {
     private HomePage homePage;
+    
+    public void initiateHomePage() {
+        homePage = new HomePage(page);
+        assertPage(homePage.waitForPage());
+    }
 
     @Test
-    @Category(org.openmrs.reference.groups.BuildTests.class)
-    public void verifyModulesAvailableOnHomePage() throws Exception {
-        homePage = new HomePage(page);
-        assertPage(homePage);
+    @Category(BuildTests.class)
+    public void verifyModulesAvailableOnHomePage() {
+        initiateHomePage();
         assertTrue(homePage.isFindAPatientAppPresent());
         assertTrue(homePage.isActiveVisitsAppPresent());
         assertTrue(homePage.isAppointmentSchedulingAppPresent());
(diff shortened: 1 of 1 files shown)


sherrif10:
This is interesting, Qa server is up and running, the test pass locally even on build but wondering why this still breaks

kdaud:
Add `@Category(BuildTests.class)` annotation to test methods which needs it.

sherrif10:
Thanks @kdaud

kdaud:
My last review on this commit would be:
Removing the redundant line `LN-80`  and
adding an empty code line after `LN-19` for quality assurance

sherrif10:
Thanks i think we can leave one Line btn L-17 And L-19

kdaud:
> i think we can leave one Line

The `line code` am talking about is after the TestClass declaration. Did you get a change to take a look at the artifacts collected [here](https://github.com/openmrs/openmrs-distro-referenceapplication/pull/476#issuecomment-870471982) ?

sherrif10:
Resolved thanks @kdaud

kdaud:
> The `line code` am talking about is after the TestClass declaration.

@sherrif10 do you mind revisiting this caption and compare with your recent commit ?

sherrif10:
You are saying after a method Test, there should be am empty  line !!

ui-tests/src/test/java/org/openmrs/reference/LoginTest.java
```diff
@@ -9,22 +9,26 @@
...
 import org.openmrs.reference.page.HomePage;
 
 import static org.junit.Assert.assertTrue;
 
-@Ignore
+
 public class LoginTest extends ReferenceApplicationTestBase {
     private HomePage homePage;
+    
+    public void initiateHomePage() {
```
k-joseph, line 23: any reasons for making this public?

k-joseph:
So this is adding no change at all but rather refactoring code, any reasons why this was initially failing if it's really passing on all, am sure bamboo would grab the issues. check through my attached review and get back to me on this here

ui-tests/src/test/java/org/openmrs/reference/LoginTest.java
sherrif10: This allows any other class to be visible to all other classes

sherrif10:
The different with this commit is that we are waiting for pages at the same time refactoring, why it was breaking on firefox is because some pages took time to load hence throwing unreachable Exceptions, we can as well see what bamboo reports

ui-tests/src/test/java/org/openmrs/reference/LoginTest.java
kdaud: When a method is consumed by functions within the same class, making it private is the best alternative. On the other hand  when a method is invoked by functions outside the host class then making it public or protected overrides private modifier. I find this [article](https://stackoverflow.com/questions/215497/what-is-the-difference-between-public-protected-package-private-and-private-in) helpful.

kdaud:
> why it was breaking on firefox is because some pages took time to load hence throwing unreachable Exceptions

Which specific pages are these ?

sherrif10:
Sorry For Late response, so within each single execution of these test, the driver is supposed to hold on and confirm whether the pages/apps are the exact and present on homePage since this test is validating the presence of apps on homePage, waiting for them gives the driver enough time to verify them

ui-tests/src/test/java/org/openmrs/reference/LoginTest.java
sherrif10: /This method should be generic , Making private would make it not  accessible /.reusable. am assuming that we can as well call this method even out side this calss cc @k-joseph  @kdaud  , i think we might need it to be  public such it can easily be called even out side this class, your suggestions will be appreciated

sherrif10:
Thanks @k-joseph

PR #489, opened 2021-07-02
RATEST-145: Fix end visit test

kdaud:
Ticket [ID](https://issues.openmrs.org/browse/RATEST-145)
Description -> Fix end visit test

### DIFF
--- a/ui-tests/src/test/java/org/openmrs/reference/EndVisitTest.java
+++ b/ui-tests/src/test/java/org/openmrs/reference/EndVisitTest.java
@@ -11,7 +11,6 @@ package org.openmrs.reference;
 
 import org.junit.After;
 import org.junit.Before;
-import org.junit.Ignore;
 import org.junit.Test;
 import org.junit.experimental.categories.Category;
 import org.openmrs.reference.groups.BuildTests;
@@ -33,14 +32,15 @@ public class EndVisitTest extends LocationSensitiveApplicationTestBase {
     }
 
     @Test
-    @Ignore
     @Category(BuildTests.class)
-    public void EndVisitTest() throws Exception {
+    public void endVisitTest() {
         FindPatientPage findPatientPage = homePage.goToFindPatientRecord();
         findPatientPage.enterPatient(patient.identifier);
+        findPatientPage.waitForPageToLoad();
         ClinicianFacingPatientDashboardPage clinicianFacingPatientDashboardPage = findPatientPage.clickOnFirstPatient();
         PatientVisitsDashboardPage patientVisitsDashboardPage = clinicianFacingPatientDashboardPage.goToRecentVisits();
         patientVisitsDashboardPage = patientVisitsDashboardPage.endVisit();
+        patientVisitsDashboardPage.waitForPageToLoad();
         assertNull(patientVisitsDashboardPage.getActiveVisit());
     }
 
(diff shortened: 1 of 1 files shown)


ui-tests/src/test/java/org/openmrs/reference/EndVisitTest.java
```diff
@@ -33,20 +32,20 @@ public void setUp() throws Exception {
...
     @Category(BuildTests.class)
-    public void EndVisitTest() throws Exception {
+    public void endVisitTest() {
         FindPatientPage findPatientPage = homePage.goToFindPatientRecord();
         findPatientPage.enterPatient(patient.identifier);
+        findPatientPage.waitForPageToLoad();
         ClinicianFacingPatientDashboardPage clinicianFacingPatientDashboardPage = findPatientPage.clickOnFirstPatient();
         PatientVisitsDashboardPage patientVisitsDashboardPage = clinicianFacingPatientDashboardPage.goToRecentVisits();
         patientVisitsDashboardPage = patientVisitsDashboardPage.endVisit();
+        patientVisitsDashboardPage.waitForPageToLoad();
```
kdaud, line 43: `Waiting the page to load` ensures that the `endVist()` operation is effective otherwise by the time `assertNull()` applies its logic to finds weather the active visit is `null` unfortunately finds it still active due to delay of the `endVisit()` operation on the visit

ui-tests/src/test/java/org/openmrs/reference/EndVisitTest.java
```diff
@@ -33,20 +32,20 @@ public void setUp() throws Exception {
     }
 
     @Test
-    @Ignore
     @Category(BuildTests.class)
-    public void EndVisitTest() throws Exception {
+    public void endVisitTest() {
         FindPatientPage findPatientPage = homePage.goToFindPatientRecord();
         findPatientPage.enterPatient(patient.identifier);
+        findPatientPage.waitForPageToLoad();
```
kdaud, line 39: `Waiting the page to load` ensures that the searched patient identifier matches with the created test patient identifier. Otherwise without` waiting the page to load`, some times the comparison is made between the searched patient identifier with any identifier (`randomly picked from the db pool`) of the existing patient data present in the  `db server instance` because the `search patient indexing criteria` that searches existing patient takes a bit longer to retrieve the `created test patient data` from the `db pool`

ui-tests/src/test/java/org/openmrs/reference/EndVisitTest.java
k-joseph: when would you use `waitForPage()` vs `waitForPageToLoad()`?

kdaud: > when would you use `waitForPage()` vs `waitForPageToLoad()`?

I would use [waitforPage()](https://github.com/openmrs/openmrs-contrib-uitestframework/blob/e3293792449c062b4934f0c29db8cb02f9156613/src/main/java/org/openmrs/uitestframework/page/Page.java#L119) when the page takes less than 500 milliseconds to load since the method implementation extends a default `TimeUnit `
(500 milliseconds) from the selenium API. However, [waitforPageToLoad()](https://github.com/openmrs/openmrs-contrib-uitestframework/blob/e3293792449c062b4934f0c29db8cb02f9156613/src/main/java/org/openmrs/uitestframework/page/Page.java#L165) is customized to a `TimeUnit` of 1000 milliseconds and this gives a room for pages that take longer than 500 milliseconds to load to rather implement `waitforPageToLoad()` over` waitforPage()`.

PR #495, opened 2021-07-08
RATEST-135: Fix add user test

kdaud:
Ticket [ID](https://issues.openmrs.org/browse/RATEST-135)
Description -> Fix add user test

### DIFF
--- a/ui-tests/src/test/java/org/openmrs/reference/AddUserTest.java
+++ b/ui-tests/src/test/java/org/openmrs/reference/AddUserTest.java
@@ -10,18 +10,25 @@
 
 package org.openmrs.reference;
 
-import org.junit.Ignore;
-import org.openmrs.reference.page.*;
-
 import java.util.List;
+
+import org.junit.Test;
+import org.junit.experimental.categories.Category;
+import org.openmrs.reference.groups.BuildTests;
+import org.openmrs.reference.page.AddEditUserPage;
+import org.openmrs.reference.page.AdministrationPage;
+import org.openmrs.reference.page.HomePage;
+import org.openmrs.reference.page.ManageUserPage;
+import org.openmrs.reference.page.SystemAdministrationPage;
 
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertTrue;
 
 public class AddUserTest extends ReferenceApplicationTestBase {
 
-    @Ignore
-    public void addUserTest() throws Exception {
+    @Test
+    @Category(BuildTests.class)
+    public void addUserTest() {
         SystemAdministrationPage systemAdministrationPage = homePage.goToSystemAdministrationPage();
         AdministrationPage administrationPage = systemAdministrationPage.goToAdvancedAdministration();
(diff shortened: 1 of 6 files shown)


ui-tests/src/test/java/org/openmrs/reference/AddUserTest.java
```diff
@@ -55,7 +62,8 @@ public void addUserTest() throws Exception {
             assertTrue(manageUserPage.getUserSavedNotification().contains("User Saved"));
 
             homePage = new HomePage(goToLoginPage().login("super_nurse", "Nurse123"));
-            homePage.getLoggedUsername().contains("super_nurse");
+            homePage.waitForPage();
```
kdaud, line 65: The home page takes a bit longer to load after login, waiting for page resolves the issue

ui-tests/src/test/java/org/openmrs/reference/AddUserTest.java
```diff
@@ -55,7 +62,8 @@ public void addUserTest() throws Exception {
...
+            assertTrue(homePage.containsText("super_nurse"));
```
kdaud, line 66: After trying different element attributes for getting the `loggedUsername`, all the element property values (`css, id, linkText, xpath` )  break only with chrome `headless=true`. Its from this reason that we rather assert the presence of the text `super_nurse` on the page since its equivalent to  `homePage.getLoggedUsername().contains("super_nurse")`

ui-tests/src/test/java/org/openmrs/reference/page/AdministrationPage.java
```diff
@@ -15,7 +15,7 @@
 public class AdministrationPage extends Page {
 
     private static final String URL = "/admin/index.htm";
-    private static final By MANAGE_USERS = By.cssSelector("#legacyui-manageUsers a");
+    private static final By MANAGE_USERS = By.cssSelector("#content a[href='/openmrs/admin/users/users.list']");
```
kdaud, line 18: The current set property would be one of the best option among the alternatives. However, the element `attribute property value` is only true when inspected via qa instance, and inspecting it via `demo.openmrs.org` & also on the` sdk local instance` the set element `property value` is false. In those two instances have mentioned the element `value property` is set to `#legacyui-mangeUsers a`. For code sustainability and quality assurance, I opted to set the property value to the suggested one since it hardly breaks with any server instance either.

kdaud:
The commit with id `db24087` might seem to be out of the scope But felt important to refactor the code base in those java classes that were overlooked within in this [ticket](https://issues.openmrs.org/browse/RATEST-160) for quality assurance.

ui-tests/src/test/java/org/openmrs/reference/AddUserTest.java
```diff
@@ -10,18 +10,25 @@
 
 package org.openmrs.reference;
 
-import org.junit.Ignore;
-import org.openmrs.reference.page.*;
```
kdaud, line 14: For quality assurance we rather explicitly import classes since `import org.openmrs.reference.page.*` would avail all the pages which is not really true with the test class in context

PR #512, opened 2021-07-26
Followup commit to handle logout visibility Exceptions on CI

sherrif10:
Follow up commit that fixes visibility of logout Element 

### DIFF
--- a/ui-tests/src/test/java/org/openmrs/reference/FormTest.java
+++ b/ui-tests/src/test/java/org/openmrs/reference/FormTest.java
@@ -1,5 +1,7 @@
 package org.openmrs.reference;
 
+import static org.junit.Assert.assertNotNull;
+
 import org.junit.After;
 import org.junit.Before;
 import org.junit.Test;
@@ -8,14 +10,13 @@ import org.openmrs.reference.groups.BuildTests;
 import org.openmrs.reference.page.AdministrationPage;
 import org.openmrs.reference.page.ClinicianFacingPatientDashboardPage;
 import org.openmrs.reference.page.HeaderPage;
-import org.openmrs.reference.page.HtmlFormsPage;
-import org.openmrs.reference.page.ManageHtmlFormsPage;
 import org.openmrs.reference.page.HomePage;
+import org.openmrs.reference.page.HtmlFormsPage;
 import org.openmrs.reference.page.ManageFormsPage;
-import org.openqa.selenium.By;
+import org.openmrs.reference.page.ManageHtmlFormsPage;
 import org.openmrs.uitestframework.test.TestBase;
+import org.openqa.selenium.By;
 
-import static org.junit.Assert.assertNotNull;
 
 /**
  * Created by nata on 24.06.15.
@@ -110,6 +111,7 @@ public class FormTest extends TestBase {
 	public void tearDown() throws Exception {
 		if (headerPage != null) {
 			headerPage.clickOnHomeIcon();
+			headerPage.waitForPageToLoad();
 			headerPage.logOut();
 		}
 	}
--- a/ui-tests/src/test/java/org/openmrs/reference/page/HeaderPage.java
+++ b/ui-tests/src/test/java/org/openmrs/reference/page/HeaderPage.java
@@ -9,7 +9,7 @@ public class HeaderPage extends Page {
 
     static final String HOME_ICON = "logo";
     static final String HOME_LINK_TEXT = "Home";
-    static final By LOGOUT = By.className("logout");
+    static final By LOGOUT = By.cssSelector(".logout > a");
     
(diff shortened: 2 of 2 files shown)


ibacher:
@sherrif10 This PR seems to break the build. Could you look into that?

sherrif10:
This PR should track the  fix the breakage on the ci build , after having this merged am sure the ci will build perfect

ui-tests/src/test/java/org/openmrs/reference/page/HeaderPage.java
```diff
@@ -9,7 +9,7 @@
 
     static final String HOME_ICON = "logo";
     static final String HOME_LINK_TEXT = "Home";
-    static final By LOGOUT = By.className("logout");
+    static final By LOGOUT = By.cssSelector("#navbarSupportedContent > ul > li.nav-item.logout > a");
```
ibacher, line 12: Why have this deeply nested a selector? Wouldn't `li.logout > a` or even just `.logout > a` do the job?

sherrif10: Sure thanks.

ui-tests/src/test/java/org/openmrs/reference/FormTest.java
```diff
@@ -1,5 +1,7 @@
 package org.openmrs.reference;
 
+import static org.junit.Assert.assertNotNull;
```
kdaud, line 3: Organizing the order of imports to follow the OpenMRS `imports convention` is a good idea to ensure quality within our code base. @sherrif10 cross check with the existing files to see the import order.

sherrif10: Hello @kdaud , I  think this is the right way imports should be organised

sherrif10:
cc @ibacher

ui-tests/src/test/java/org/openmrs/reference/FormTest.java
sherrif10: Uhmm But when i save still this import remains on top , i dont understand by saying following the alphabetic order, imports should be well formatted and organised by the ide. i dont see any problem with that  cc @ibacher

kdaud: > Uhmm But when i save still this import remains on top , i dont understand by saying following the alphabetic order, imports should be well formatted and organised by the ide

Some configuration within the IDE results into this! However, this can be handled with ease when the IDE is well configured!

>  i dont see any problem with that 

BTW this is not a bug and has no harm at all with other existing code base. However, this means we are not ensuring standards within our code base with is not a good idea

sherrif10: My ide is well configured and still i think this is the right way,i actually dont mind about the right way however my ide is well configured to automatically organise import statements, i can say the first commit which was merged that first authores same imports where the later was first also had no issus. it doesnot matter which is first which is last, what matters is whether your imports are well organised. what i mean is 

There is a difference btn  
i**mport org.openqa.selenium.support.ui.ExpectedConditions;
import org.openqa.selenium.support.ui.Select;
import org.openqa.selenium.support.ui.WebDriverWait;

import java.net.MalformedURLException;
import java.net.URL;
import java.time.Duration;**

And 
**import org.openqa.selenium.support.ui.ExpectedConditions;
import org.openqa.selenium.support.ui.Select;
import org.openqa.selenium.support.ui.WebDriverWait;
import java.net.MalformedURLException;
import java.net.URL;
import java.time.Duration;**

And there is no difference btn

import org.openqa.selenium.support.ui.ExpectedConditions;
import org.openqa.selenium.support.ui.Select;
import org.openqa.selenium.support.ui.WebDriverWait;
import java.net.MalformedURLException;
import java.net.URL;
import java.time.Duration;

and 
import org.openqa.selenium.support.ui.ExpectedConditions;
import org.openqa.selenium.support.ui.Select;
import org.openqa.selenium.support.ui.WebDriverWait;

**import java.net.MalformedURLException;
import java.net.URL;
import java.time.Duration;**

The only difference is that the second last is not well organised which is not allowed as openmrs conventions , So according to your suggestions as if the first should come last which is not a strong decision to lie on , And if i would have time i would have shown you different code that have some thing similar.  cc @ibacher  whats your suggestion

sherrif10:
Hey @ibacher  Kindly review thanks , Hope this is ready

ui-tests/src/test/java/org/openmrs/reference/FormTest.java
ibacher: Static imports (like this one) always come before standard imports (note `import static` rather than just `import`), so this is actually inline with our coding standards.

PR #519, opened 2021-08-01
RATEST-90:Add Patient Appointment Test

insookwa:
https://issues.openmrs.org/browse/RATEST-90?filter=-1
Added a few methods to make the test run all day of the week. kindly review @kdaud  @sherrif10  @dkayiwa

### DIFF
--- a/ui-tests/src/test/java/org/openmrs/reference/AddPatientAppointmentTest.java
+++ b/ui-tests/src/test/java/org/openmrs/reference/AddPatientAppointmentTest.java
@@ -10,64 +10,82 @@
 
 package org.openmrs.reference;
 
+import static org.junit.Assert.assertTrue;
+
 import org.junit.After;
 import org.junit.Before;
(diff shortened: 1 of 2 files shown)


kdaud:
@insookwa please ensure that only your changes are included in the commit otherwise am seeing existing codes being reflected as your changes which is not true. You might need to do some configurations in your IDE.

insookwa:
> @insookwa please ensure that only your changes are included in the commit otherwise am seeing existing codes being reflected as your changes which is not true. You might need to do some configurations in your IDE.

i think it appears like this because after work i ctrl A and then format . in order to have the OpenMRS format

kdaud:
> i think it appears like this because after work i ctrl A and then format

Are you sure all the changes within the commit belong to you? I mean ensuring that whatever git is tracking as a change was introduced by you. This quickens the reviewers time on the your PR!

PR #520, opened 2021-08-03
RATEST-157: Adding page functionality for manage user test workflow

kdaud:
Ticket [ID](https://issues.openmrs.org/browse/RATEST-157)
Description -> Adding page functionality for manage user account test workflow

### DIFF
--- a/ui-tests/src/test/java/org/openmrs/reference/page/ManageUserAccountPage.java
+++ b/ui-tests/src/test/java/org/openmrs/reference/page/ManageUserAccountPage.java
@@ -1,0 +1,148 @@
+package org.openmrs.reference.page;
+
+import java.util.List;
+
+import org.openmrs.uitestframework.page.Page;
+import org.openqa.selenium.By;
+
+public class ManageUserAccountPage extends Page {
+
+    public static String URL = "/adminui/systemadmin/accounts/manageAccounts.page";
+    private static final By SEARCH_USER = By.cssSelector("#list-accounts_filter input[type=text]");
+    private static final By ADD_NEW_ACCOUNT = By.cssSelector("#content a[href='/openmrs/adminui/systemadmin/accounts/account.page']");
+    private static final By FAMILY_NAME = By.cssSelector("#adminui-familyName-field");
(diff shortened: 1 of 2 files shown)


HerbertYiga:
looking good to me cc @ibacher @brandones

ibacher:
@kdaud Do we have any tests exercising this new page functionality?

ui-tests/src/test/java/org/openmrs/reference/page/ManageUserAccountPage.java
```diff
@@ -0,0 +1,148 @@
...
+        clickOn(GENDER_FEMALE);
+    }
+
+    public void clickOnAddUserAccount() {
+        clickOn(ADD_USER_ACCOUNT);
+    }
+   
+    public void enterUsername(String username) {
+        findElement(USER_NAME).clear();
+        findElement(USER_NAME).sendKeys(username);
```
ibacher, line 70: I'm of two minds about this pattern. On the one hand, its necessary to reliably do what the test needs to do. On the other hand, if there's data pre-populated into that field, that seems like an issue that should be flagged. Thoughts?

kdaud:
> Do we have any tests exercising this new page functionality?

The test [here](https://issues.openmrs.org/browse/RATEST-157) is relying on the page functionality in this commit.

ui-tests/src/test/java/org/openmrs/reference/page/ManageUserAccountPage.java
kdaud: Here is my thought on the epic:
Reliably making the pattern to directly do what the logic is supposed to perform would be ok, however ,on the the other hand there could be data pre-populated in the field and invoking  clear() method makes things right using the same pattern. 
For-example when editing user account details the same pattern will cater for such scenarios than again writing another pattern that can edit the data, so I thought catching two birds in one net would be cool than using two nets to catch two birds. 

Echoing on the first thought invoking clear() method has no harm at all but does the magic for the second  thought!

ibacher (approved):

ibacher:
Thanks @kdaud. This LGTM.

PR #551, opened 2021-09-01
RATEST-155: Resurrect addRolesToUser test

kdaud:
Ticket [ID](https://issues.openmrs.org/browse/RATEST-155)

Decription -> Resurrect addRolesToUser test

### DIFF
--- a/ui-tests/src/test/java/org/openmrs/reference/AddRolesToUserTest.java
+++ b/ui-tests/src/test/java/org/openmrs/reference/AddRolesToUserTest.java
@@ -1,20 +1,21 @@
 package org.openmrs.reference;
 
-import org.junit.After;
-import org.junit.Ignore;
-import org.openmrs.reference.page.AdministrationPage;
-import org.openmrs.reference.page.ManageUserPage;
-
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Random;
 
+import org.junit.After;
+import org.junit.experimental.categories.Category;
+import org.junit.jupiter.api.Test;
+import org.openmrs.reference.groups.BuildTests;
+import org.openmrs.reference.page.AdministrationPage;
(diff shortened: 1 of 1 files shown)


kdaud:
While investigating AddRolesToUser test have noticed that the number of of accessible applications by SystemDeveloper does not match those that is being recorded by the test [here](https://github.com/openmrs/openmrs-distro-referenceapplication/blob/b03dc85483d211c712025364cdda8ace88b8e347/ui-tests/src/test/java/org/openmrs/reference/AddRolesToUserTest.java#L59).
 According to the [logs](https://pastebin.com/Ji57eJhU) the issue is being cited as` java.lang.AssertionError: role roleStrings.SystemDeveloper=10 doesn't have matching number of accessible applications: should be:10is:9`. 
Changing the integer value from 10 to 9 the test still fails with the these [logs](https://pastebin.com/MQZmK1JA) which I think could be coming from the back end which might not be easy to fix it from the front end.

Removing these code block [here ](https://github.com/openmrs/openmrs-distro-referenceapplication/blob/b03dc85483d211c712025364cdda8ace88b8e347/ui-tests/src/test/java/org/openmrs/reference/AddRolesToUserTest.java#L45) within the test passed the test however, the code block is so useful to crosscheck whether `homePage.numberOfAppsPresent() != role.getValue()` and if the boolean value returns true the it throws an `AssertionError`!

@sherrif10 any recommendation on this test and how we can make it pass the build plan ?

kdaud:
@sherrif10 did you take a look at the comment [here](https://github.com/openmrs/openmrs-distro-referenceapplication/pull/551#issuecomment-911820694) ?

sherrif10:
> whether the `homePage.numberOfAppsPresent() != role.getValue()` and if the boolean value returns true then it should throw an `AssertionError`!

How about homePage.numberOfAppsPresent() ==role.getValue(); does this also fails the test.

PR #561, opened 2021-09-14
update released versions

HerbertYiga:
Updating released versions

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -34,13 +34,13 @@
 		<uiframeworkVersion>3.21.0</uiframeworkVersion>
 		<registrationcoreVersion>1.11.0</registrationcoreVersion>
 		<registrationappVersion>1.24.0</registrationappVersion>
-		<eventVersion>2.9.0</eventVersion>
+		<eventVersion>2.10.0</eventVersion>
 		<idgenVersion>4.7.0</idgenVersion>
 		<emrapiVersion>1.32.0</emrapiVersion>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -53,7 +53,7 @@
 		<htmlformentryVersion>4.1.0</htmlformentryVersion>
 		<htmlformentryuiVersion>2.0.0</htmlformentryuiVersion>
 		<coreappsVersion>1.33.0</coreappsVersion>
-		<webservices.restVersion>2.32.0-SNAPSHOT</webservices.restVersion>
+		<webservices.restVersion>2.31.0</webservices.restVersion>
```
HerbertYiga, line 56: hi @dkayiwa  how does this pr look to you

HerbertYiga: > is there any reason why you did not just release version 2.32.0 of the module?

@dkayiwa  its because i had already released version 2.31.0!

dkayiwa: @HerbertYiga it has been a while since that was done. And a number of strategic commits have been done to that module and a few others. My suggestion would be going through them again and see if even more modules should be released and included in the upcoming reference application.

HerbertYiga: > it has been a while since that was done. And a number of strategic commits have been done to that module and a few others. My suggestion would be going through them again and see if even more modules should be released and included in the upcoming reference application.

well noted

HerbertYiga:
moved this here https://github.com/openmrs/openmrs-distro-referenceapplication/pull/567

PR #580, opened 2021-10-04
clickOnCurrentDay -> clickOnNextWeekday

ibacher:


dkayiwa:
@ibacher isn't simpler and more logical to set the true value for includeWeekend, right within the test?

ibacher:
@dkayiwa I actually think this is the better solution for the Selenium tests. Otherwise, we'd have to either:

1. Write a flow to set the GP through the UI
2. Make a REST service call

Which makes the success of this test dependent on those ones. And in either of those cases, the test could still fail because of gremlins, e.g., if another test run adjusts the setting on the QA server while the test is being run.

I'm not opposed to the idea. I just don't think I could implement it in 5 minutes 😁.

dkayiwa:
Sounds fair enough. :)
Does it help to add a comment explaining why we are using clickOnNextWeekday?

sherrif10:
Thanks  alot @ibacher . So ideally this  will override functionality of picking days before Friday without manually setting the GP in UI!

PR #609, opened 2022-01-26
Temporarily switch platform from 2.6.0 SNAPSHOT to 2.4.3

kdaud:
For one reason or the other,  switching qa server to run on the latest platform snapshot is causing all the automated tests to fail both on github actions and Bamboo.  I think for now we can have the server run against 2.4.3 version.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -28,7 +28,7 @@
 
 	<properties>
 		<project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-		<openMRSVersion>2.6.0-SNAPSHOT</openMRSVersion>
+		<openMRSVersion>2.4.3</openMRSVersion>
 		<referencemetadataVersion>2.12.0</referencemetadataVersion>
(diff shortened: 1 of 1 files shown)


dkayiwa:
@kdaud have you tried to look at the error logs to see the actual failures? Because we eventually need to anyway fix the tests because this is the platform version that we are releasing next.

sherrif10:
@kdaud The actual failure is that Person/patient is failing to get posted and the fix should be handled from our restclient class or testdata classes

kdaud:
> @kdaud The actual failure is that Person/patient is failing to get posted and the fix should be handled from our restclient class or testdata classes


@sherrif10 do you mind to share how the fix would look like?

sherrif10:
@kdaud  Working on the fix.

kdaud:
> @kdaud Working on the fix.

@sherrif10 any updates?

sherrif10:
Am looking into the logs,probably as soon as possible.

PR #637, opened 2022-09-29
(chore) Add Fast Data Entry and Cohort Builder to spa-build-config

ZacButko:
This adds @openmrs/esm-fast-data-entry-app and @openmrs/esm-cohort-builder-app to spa-build-config.json so that they will be included in the 3.x ref app

Note that as of writing the cohort-builder-app is not registered on npmjs.com or any other package registry

### DIFF
--- a/frontend/spa-build-config.json
+++ b/frontend/spa-build-config.json
@@ -28,7 +28,9 @@
     "@openmrs/esm-patient-registration-app": "next",
     "@openmrs/esm-patient-search-app": "next",
     "@openmrs/esm-openconceptlab-app": "next",
-    "@openmrs/esm-dispensing-app": "next"
+    "@openmrs/esm-dispensing-app": "next",
+    "@openmrs/esm-fast-data-entry-app": "next",
+    "@openmrs/esm-cohort-builder-app": "next"
   },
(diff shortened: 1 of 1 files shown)


ibacher:
Ok... this reminds me why this wasn't here in the first place. We need to get the cohort-builder-app published to NPMJS before this can be safely merged, otherwise the build process will fail. Apart from that, this LGTM.

anjula-sack:
> Ok... this reminds me why this wasn't here in the first place. We need to get the cohort-builder-app published to NPMJS before this can be safely merged, otherwise the build process will fail. Apart from that, this LGTM.

I updated the workflow https://github.com/openmrs/openmrs-esm-cohortbuilder/pull/27 after this it will get published right?

anjula-sack:
> Ok... this reminds me why this wasn't here in the first place. We need to get the cohort-builder-app published to NPMJS before this can be safely merged, otherwise the build process will fail. Apart from that, this LGTM.

This can be merged. cohort builder is now [published](https://www.npmjs.com/package/@openmrs/esm-cohort-builder-app).

ibacher (approved):

PR #640, opened 2022-10-04
RA-1993: lab_tests-core-demo.zip to be self contained and able to load in offline mode

Ruhanga:
Followup on Ticket: https://issues.openmrs.org/browse/RA-1993


ibacher:
Does this duplicate concepts from the test CSV? If so, could we delete that file too?

Ruhanga (commented):
Thanks for highlighting this @ibacher, I've deleted duplicate entries, that for Haemoglobin and Hematocrit. Could you review again?

PR #642, opened 2022-10-06
Create relationshiptypes.csv

gracepotma:
It's been bothering me that we have very few relationship examples in our Demo registration experience - e.g. you can't record a spouse relationship, and you're out of luck if the relationship type isn't listed (there was no "other" option). I wanted to merge this right away but figured it should probably wait for the next release and I wanted to double check it with folks like @ibacher & @vasharma05 first.


ibacher (approved):

ibacher:
To be fair, I'm familiar with orgs that bypass our more formal relationship modelling to just have, e.g., person attributes to capture relevant relationship information, but I don't see an issue with having more options!

Ruhanga (commented):
Thanks for adding in these @gracepotma. A minor correction, this file should have gone in a `relationshiptypes ` folder. For example on the path `distro/configuration/relationshiptypes/relationshiptypes.csv`

PR #644, opened 2022-10-07
 O3-1557: run e2e tests on the 3.x branch at openmrs-distro-reference application

jnsereko:
I strongly stand by us running all tests rather than smoke tests.  I have honestly failed to sieve out what test is more worth than the other.  

I will also create another workflow to run smoke tests so devs can optionally skip (ignore) the all e2e tests workflow in case they are overwhelmed by the reds while merging. 

cc @dkayiwa @gracepotma @ibacher

### DIFF
--- a/.github/workflows/refapp-3x-vitals-and-biometrics.yml
+++ b/.github/workflows/refapp-3x-vitals-and-biometrics.yml
@@ -33,4 +33,4 @@ jobs:
         uses: actions/upload-artifact@v2
         with:
           name: Screen recordings of failed tests
-          path: packages/cypress/cypress/videos
+          path: openmrs-test-3refapp/packages/cypress/cypress/videos
--- a/.github/workflows/refapp-3x-vitals-and-biometrics.yml
+++ b/.github/workflows/refapp-3x-vitals-and-biometrics.yml
@@ -25,7 +25,7 @@ jobs:
           repository: openmrs/openmrs-test-3refapp
           path: openmrs-test-3refapp
       - name: Installing Repo Dependencies
-        run: cd openmrs-test-3refapp && yarn install
+        run: cd openmrs-test-3refapp && yarn install --immutable
       - name: Run e2e tests
         run: cd openmrs-test-3refapp/packages/cypress && yarn refappCypress
       - name: Upload screen recordings of failed tests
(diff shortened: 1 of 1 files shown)


.github/workflows/refapp-3x-vitals-and-biometrics.yml
```diff
@@ -0,0 +1,36 @@
...
+        run: docker compose up -d --build
+      - name: Wait for Openmrs instance to start
+        run: while [[ "$(curl -s -o /dev/null -w ''%{http_code}'' http://localhost:8080/openmrs/login.htm)" != "200" ]]; do sleep 10; done
+      - name: Clone tests
+        uses: actions/checkout@v3
+        with:
+          repository: openmrs/openmrs-test-3refapp
+          path: openmrs-test-3refapp
+      - name: Installing Repo Dependencies
+        run: cd openmrs-test-3refapp && yarn install
```
ibacher, line 28:
```suggestion
        run: cd openmrs-test-3refapp && yarn install --immutable
```

.github/workflows/refapp-3x-vitals-and-biometrics.yml
```diff
@@ -0,0 +1,36 @@
...
+      - name: Run e2e tests
+        run: cd openmrs-test-3refapp/packages/cypress && yarn refappCypress
+      - name: Upload screen recordings of failed tests
+        if: ${{ failure() }}
+        uses: actions/upload-artifact@v2
+        with:
+          name: Screen recordings of failed tests
+          path: packages/cypress/cypress/videos
```
ibacher, line 36:
```suggestion
          path: openmrs-test-3refapp/packages/cypress/cypress/videos
```

ibacher (changes requested):
I think this PR is mis-targetted. In particular, this should run as part of the Bamboo job rather than GitHub Actions here. Triggering this as part of GitHub Actions means it will only run on PRs and merges _to this repo_. Most such merges are adding metadata to the RefApp and these tests add very little value there. However, the Bamboo job is currently triggered on every commit to the RefApp frontend repo, so it will be able to catch breaking changes after they are committed.

PS Thanks @jnsereko for all your work on this. It's really just a matter of making sure we run these tests at the right point.

jnsereko:
thank you so much for the response @ibacher 
I kindly request you and @dkayiwa to help me an we follow up on this at https://openmrs.slack.com/archives/CP343JH1R/p1665404314967979

jayasanka-sack:
Closing this PR as it has already been addressed in https://issues.openmrs.org/browse/O3-2493

PR #647, opened 2022-10-17
O3-858: Adding a demo lab order entry form.

Ruhanga:
Ticket: https://issues.openmrs.org/browse/O3-858


mks-d (commented):
Thanks @Ruhanga!

Shouldn't there be some extra config somewhere else that makes this form show in the O3 UI? Or that just happens?

Ruhanga (commented):
Hi @mks-d, adding the form via iniz ampath forms should suffice.

ibacher:
This seems to generate an ObsGroup instead of actual orders... while I understand this is a way many implementations have handled orders, doesn't this imply we need some kind of backend handling to turn these observations into actual orders? Otherwise, we're going to have to build a weird double logic into the frontend to display actual "Order" orders and these "Obs" orders...

reagan-meant (approved):

Ruhanga:
The obsGroup is simply a ui grouping and not persisted. No actual obs is persisted by the underlying widget.

ibacher:
Thanks @Ruhanga! I get it now.

PR #653, opened 2022-10-27
Fix for CIEL-Frequencies concepts to resolve errors

gracepotma:
Version 1 of the CIEL-Frequencies metadata was triggering the following errors: 
image

I **removed the concept 164844** ("HIV status of source person") which is the source of both these issues and is not necessary for this collection. HIV Status really doesn't have anything to do with the frequency of taking a drug so I'm not sure how that concept got added here at all. 

(FYI @rbuisson since it seems you have been active on this file too.)


ibacher (approved):

mks-d:
@gracepotma @ibacher in Ozone there are a number of pieces of metadata for which we keep the configurations in sync manually between OpenMRS and the pharmacy/inventory system (namely Odoo in this case). This change has broken this alignment between systems and we need to understand what was changed exactly. There is no ticket here so there is no documentation as to what this PR intends to fix and why. I'm sure there were good reasons, but it's off process and undocumented.

Cc @basrem @nuhuhmutebi could you drill down and report back?

ibacher:
@mks-d The change is what Grace put in the description of the PR, i.e., one concept was removed from the list of references.

ibacher:
@mks-d The delete-and-replace is so we can track which version the zips correspond to on OCL... While that's derivable from the JSON in the ZIP, it's easier if it's on the file. The change described is the change from version 1 to version 3 of that concept source.

You're right that there were other things that changed. Looking into V1 of this file, it included no concept mappings at all, which is... not correct. However, that seems to be because [the previous collection used](https://app.openconceptlab.org/#/users/mksrom/collections/CIEL-frequencies/HEAD/expansions/autoexpand-HEAD/concepts/) was built incorrectly.

ibacher:
@mks-d Are you able to give us any kind of description of the problem this caused? Because there's clearly some issue with metadata management, but without some idea what went wrong its hard to know what needs to be fixed...

mks-d:
@ibacher sure, we are still investigating failing drug order sync cases to understand better the extent for the problem and how to report it back here in a useful way.

mks-d:
@gracepotma / @ibacher sorry about the noise here 🙏 , turns out that this commit was never an issue on our end. All good as far as we can tell.

PR #655, opened 2022-11-07
Add CIEL severities concept set

hadijahkyampeire:
Add a ZIP for the CIEL severities concept https://app.openconceptlab.org/#/orgs/CIEL/sources/CIEL/concepts/159411/

We don't have any severities concept and that is causing the `/allergy` POST endpoint to fail, I have added the zip file for the above concept so that we use UUIDs for that severity.

looking at the UUIDs we are currently using in the FE code, they are similar to the ones for the concepts in this concept set.
i.e 
```
Moderate concept: 1499AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA
Mild: 1498AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA
Severe: 1500AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA
```
which means this should fix the allergy posting failure.


denniskigen (approved):
I can't tell if there's a naming convention that these archives ought to conform to just by looking at the filenames in the `ocl` directory. That aside, this LGTM! Thanks, @hadijahkyampeire!

PR #657, opened 2022-11-15
TRUNK-6083 Updating O3 to new core image

rkorytkowski:


rkorytkowski:
@dkayiwa @ibacher  It appears I'm not able to merge without review. Are we enforcing now all commits to go through PR?

ibacher:
@rkorytkowski Yes... This is mostly about ensuring sanity for people merging metadata into the application.

docker-compose.yml
```diff
@@ -3,6 +3,7 @@ version: "3.7"
 services:
   gateway:
     image: openmrs/openmrs-reference-application-3-gateway:${TAG:-nightly}
+    build: ./gateway
```
ibacher, line 6: I'd prefer we leave these out. Since we're not using docker-compose to actually build this, this file basically serves as a reference for implementations.

docker-compose.yml
```diff
@@ -41,10 +46,15 @@ services:
   db:
     image: mariadb:10.8.2
     command: "mysqld --character-set-server=utf8 --collation-server=utf8_general_ci"
+    healthcheck:
+      test: "mysql --user=${OMRS_DB_USER:-openmrs} --password=${OMRS_DB_PASSWORD:-openmrs} --execute \"SHOW DATABASES;\""
```
ibacher, line 50: It might be better to use the `mysqladmin ping` command, no? This would fail if the openmrs user just doesn't have permissions to list the available databases.

ibacher (approved):

docker-compose.yml
rkorytkowski: mysqladmin ping sometimes gives false positive, e.g. "This is 0 even in case of an error such as Access denied, because this means that the server is running but refused the connection, which is different from the server not running. ", see https://dev.mysql.com/doc/refman/8.0/en/mysqladmin.html. The permission issue should not be the case here, because we setup the DB with this user, but if it's a concern then I can just do "SELECT 1;" or something like that. It's more bulletproof.

docker-compose.yml
rkorytkowski: Well, actually one could use docker-compose.yml to build the production image and docker-compose.override.yml to build the dev image, but it's not setup this way yet so it's fair to leave it out for now. See e.g. https://github.com/openmrs/openmrs-core/blob/master/docker-compose.override.yml#L20-L24 and https://github.com/openmrs/openmrs-core/blob/master/docker-compose.yml#L28-L30
The dev image has all sources included whereas the production image has just binaries. I can imagine we will benefit from that in O3 at some point as well.

PR #672, opened 2023-01-13
RA-1994: Double commit for 3.0.0 beta.1 release.

enyachoke:
Ticket: https://issues.openmrs.org/browse/RA-1994


distro/pom.xml
```diff
@@ -27,7 +27,7 @@
     <initializer.version>2.4.0</initializer.version>
     <fhir2.version>1.7.0</fhir2.version>
     <webservices.rest.version>2.38.0</webservices.rest.version>
-    <addresshierarchy.version>2.15.1</addresshierarchy.version>
+    <addresshierarchy.version>2.16.0-SNAPSHOT</addresshierarchy.version>
```
mks-d, line 30: @enyachoke I don't think this is necessary. Do you know why we would make an exception with AH to revert it back to its latest snapshot, when we don't do that for Queues and some other modules that are supposedly under active development?

mks-d (commented):
LGTM and ⚠️ careful ⚠️ this PR needs to be merged using '**Rebase and merge**'.

mks-d (approved):

PR #686, opened 2023-01-31
O3-1764: Added better Appointment Service type demo options

gracepotma:
Goal of this PR is to add some more reasonable-sounding options for Appointment Service field options to our Metadata, per https://issues.openmrs.org/browse/O3-1764. 
* I added General Medicine, Outpatient Department, and Rehabilitation as options, and used the code for the "General" service type. 
* I referenced this directory for the Specialty IDs: https://github.com/openmrs/openmrs-distro-referenceapplication/blob/main/distro/configuration/appointmentspecialities/specialities.csv 
* I *retained* the options for "General" and "Malnutrition" just in case these are part of someone's tests or Demo Data (want to keep the scope of this change impact extremely minimal).


hadijahkyampeire (approved):
Thanks @gracepotma let’s merge this and see if we get the desired service types.

PR #701, opened 2023-02-25
O3-1709: Permissions for which Dispensing fields can be edited and by…

mogoodrich:
… which role


mogoodrich:
fyi @ibacher ... also, any easy way I can test I got these right without setting up a full ref app instance? :)

ibacher (approved):

ibacher:
@mogoodrich Not yet, but when the Iniz validator is working again, we're going to integrate that as a check in the CI process.

PR #715, opened 2023-04-04
O3-1727: New Test Results Entry form

gracepotma:
Adds a simple but long Test Results Entry form that can be used to manually report recent lab results for some common test types.
The goal is to unblock groups like Faimer who want to be able to record test results, without hooking up a Lab Reporting System. 
Known deficiencies:
* The date/time stamp from the form SUBMISSION will be applied to the Test concept, even if this is not actually reflective of the Test RESULT TIME. We should come back to this in the near future when we work on improving Past Visits/RDE flows. 
* multiCheckbox was not working in the form builder ([logged here](https://github.com/openmrs/openmrs-esm-form-builder/issues/73)) so for now I was forced to use Radio where Checkbox would be much more appropriate. 

Issue for this PR: https://issues.openmrs.org/browse/O3-1727

### DIFF
--- a/distro/configuration/ampathforms/TestResultsEntryFormv2
+++ b/distro/configuration/ampathforms/TestResultsEntryFormv2
@@ -1,0 +1,2241 @@
+{
+  "name": "Test Results Entry Form",
+  "version": "2",
+  "published": true,
+  "retired": false,
+  "encounter": "Lab Results",
+  "pages": [
+    {
+      "label": "Hematology",
+      "sections": [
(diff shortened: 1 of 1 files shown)


ibacher (approved):

rbuisson:
Looks like this file is missing an extension. I'm guessing `.json` ;)

gracepotma:
Yup, how embarrassing. Thanks @rbuisson for the catch. Fixed in mainline.

PR #717, opened 2023-04-18
RA-2023: Making configs validated through Iniz Validator.

Ruhanga:
Ticket: https://issues.openmrs.org/browse/RA-2023
Depends on https://github.com/openmrs/openmrs-contrib-packager-maven-plugin/pull/7


distro/pom.xml
```diff
@@ -183,6 +183,20 @@
 
   <build>
     <plugins>
+      <plugin>
+        <groupId>org.openmrs.maven.plugins</groupId>
+        <artifactId>openmrs-packager-maven-plugin</artifactId>
+        <version>1.5.0-SNAPSHOT</version>
```
ibacher, line 189: We need to do a release of the Maven plugin before this can be incorporated. I think it's OK to tolerate SNAPSHOTs in the properties that are driving the distro configuration, but completely wrong to do that for Maven plugins.

distro/configuration/concepts/misc-core_data.csv
```diff
@@ -1,6 +1,4 @@
 Uuid,Void/Retire,Same as mappings,Fully specified name:en,Fully specified name:fr,Fully specified name:es,Short name:en,Short name:fr,Short name:es,Description:en,Description:fr,Description:es,Data class,Data type,Answers,_version:1,_order:1000
-1065AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA,,CIEL:1065;PIH:1065;AMPATH:1065;SNOMED CT:373066001,Yes,Oui,Sí,,,,Generic answer to a question.,,,Misc,N/A,,,
```
ibacher, line 2: Why are we removing these concepts?

distro/configuration/concepts/drugs_concepts.csv
```diff
@@ -1,14 +1,12 @@
 Uuid,Void/Retire,Short name:en,Fully specified name:en,Fully specified name:fr,Data class,Data Type,Members,_version:1,_order:500
```
ibacher, line 1: Ditto here as for drugs: the concepts need to be fixed. We shouldn't be removing them.

ibacher (commented):
Thanks @Ruhanga! This is important work, but it's also important that we not delete metadata just because it fails in the validator. We need to either fix the metadata or fix the validator.

distro/pom.xml
Ruhanga: Yes for sure. That's precisely the plan. I'd simply done this in awaiting a release of the plugin.

distro/configuration/metadatatermmappings/metadatatermmappings.csv
```diff
@@ -1,3 +0,0 @@
-Uuid,Void/Retire,Mapping code,Mapping source,Metadata class name,Metadata Uuid,_order:2000
```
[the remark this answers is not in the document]
Ruhanga: For a while these don't seem to be necessarily needed (since no one is noticing their absence in a Ref App distro) and so IMO, removing them won't do any harm I guess.

distro/configuration/concepts/misc-core_data.csv
Ruhanga: These are duplicates and already come with the OCL packages.

distro/configuration/concepts/drugs_concepts.csv
Ruhanga: Removing them sounds like the right fix since they are duplicates already brought in by the OCL packages.

distro/configuration/drugs/drugs.csv
```diff
@@ -23,21 +23,13 @@ a80acf74-a85c-4fef-8dc4-d763e364663f,,Metformin,79651AAAAAAAAAAAAAAAAAAAAAAAAAAA
 721497f0-dcf4-4f05-8fe7-e2e487b0d774,,Metformin Co 500mg,79651AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA,1513AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA, 500mg,
```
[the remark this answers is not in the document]
Ruhanga: For a while they don't load and this is not noticed by anyone needing them. IMO I'd rather remove them so they are actually added when needed.

distro/configuration/concepts/drugs_concepts.csv
ibacher: Ok... As long as their all duplicated.

distro/configuration/metadatatermmappings/metadatatermmappings.csv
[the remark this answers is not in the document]
Ruhanga: @ibacher, I've been looking into this silent failure to load the `emrapi.Metadata_Source-1.zip` and the PR below should fix it.

https://github.com/mekomsolutions/openmrs-module-initializer/pull/241

Ruhanga (commented):
Hi @ibacher, could you review and share your final remarks? Thanks. I've added github actions too. See the branch's successful build [here](https://github.com/Ruhanga/openmrs-distro-referenceapplication/tree/validation).

distro/pom.xml
Ruhanga: Done. The Plugin used is 1.7.0.

gracepotma:
@Ruhanga I'm concerned about the metadata deletions. Can you tell me why these drug concepts are being removed? 
If the problem is being caused due to those line items being empty in the strength field (e.g. not having any mg's assigned) I can fix that. Since I introduced that.

Ruhanga:
@gracepotma, at the moment the deleted drugs are as good as not present  (even before deletion). They are already failing. I would clean them up in favour of adding them on an error free configuration.

Ruhanga:
>Would you like me to, soon, go through all drugs and ensure those are in /ocl as well, for consistency?

@gracepotma you could do so at-least soon after this is merged and released?

gracepotma:
Ah gotcha, thank you Nathan. That totally makes sense then. And sure, I've ticketed this for myself here: https://issues.openmrs.org/browse/O3-2077

.github/workflows/build-test.yml
```diff
@@ -0,0 +1,27 @@
+name: Build and Validate Configuration
+
+on:
+  pull_request:
+    branches: [master]
```
ibacher, line 5:
```suggestion
    branches: [main]
```

.github/workflows/build-test.yml
```diff
@@ -0,0 +1,27 @@
...
+    runs-on: ubuntu-latest
+    strategy:
+        matrix:
+          java: [ '8' ]
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Set up JDK ${{ matrix.java }}
+        uses: actions/setup-java@v2
```
ibacher, line 21:
```suggestion
        uses: actions/setup-java@v3
```

.github/workflows/build-test.yml
```diff
@@ -0,0 +1,27 @@
...
+        with:
+          distribution: 'temurin'
+          java-version: ${{ matrix.java }}
```
ibacher, line 24:
```suggestion
          java-version: ${{ matrix.java }}
          cache: 'maven'
```

pom.xml
```diff
@@ -164,9 +164,9 @@
 
   <pluginRepositories>
     <pluginRepository>
-      <id>openmrs-repo</id>
-      <name>OpenMRS Public Repository</name>
-      <url>https://mavenrepo.openmrs.org/public</url>
+      <id>openmrs-repo-releases</id>
+      <name>OpenMRS Releases Repo</name>
+      <url>https://openmrs.jfrog.io/artifactory/releases</url>
```
ibacher, line 169: I don't think we need this change; specifically, `public` aliases all the repos in the OpenMRS Maven instance; `releases` does not.

ibacher (changes requested):
Thanks @Ruhanga! I think we need some adjustments to the POM changes and the GitHub Actions, but otherwise, I'm happy with this.

.github/workflows/build-test.yml
ibacher: We use `main` not `master` in this repo.

.github/workflows/build-test.yml
```diff
@@ -0,0 +1,27 @@
+name: Build and Validate Configuration
+
+on:
+  pull_request:
+    branches: [master]
+
+  push:
+    branches-ignore: [master]
```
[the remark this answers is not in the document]
ibacher: We need to exclude the branches for other versions. Tbh, I'd probably just drop the `push` trigger altogether.

pom.xml
```diff
@@ -116,6 +116,17 @@
   </build>
 
   <repositories>
+    <repository>
```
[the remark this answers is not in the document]
Ruhanga: Something weird's happening the OpenMRS repo. If an artefact is missing or poor redirection happens, builds fail. That's why I'm leaving it for last to prevent invalid failures.

pom.xml
Ruhanga: This is somewhat connected to [this](https://openmrs.slack.com/archives/C03U2P841/p1682617009756369?thread_ts=1682419613.318009&cid=C03U2P841) suggestion @ibacher.

Ruhanga: So what's happening is the openmrs packager plugin artefact isn't downloadable via `https://mavenrepo.openmrs.org/public`, but is via `https://openmrs.jfrog.io/artifactory/releases`. Could this be a problem with the alias url or jfrog?

Ruhanga: I think I get your point @ibacher. In  essence I should use `https://openmrs.jfrog.io/artifactory/public` instead of `https://openmrs.jfrog.io/artifactory/releases`?

pom.xml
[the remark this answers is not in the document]
mks-d: @ibacher in the end something's fishy when there's redirections to another Maven repository. In particular this happens with Iniz because it is proxied maybe? And it looks like letting the build fetch it from the original repo first solves the issue.

In short this proxying is just not working really, we should never have had to add Mekom's repos here, let alone first.

So while I agree that this is not idea altogether, it seems to finally unblock the build.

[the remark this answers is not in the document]
Ruhanga: @ibacher, where you able to disable proxying for the Mekom repo on the OpenMRS repos so I could try again re-arranging the repositories?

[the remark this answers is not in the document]
ibacher: We aren't proxying the Mekom repo in the OpenMRS repo any more... I'm not sure that we have been for a while. What's weird about this to me is that we've been successfully pulling down, e.g., Iniz with the previous ordering.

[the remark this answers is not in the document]
ibacher: So when I try to run this PR the error I see is this:

```
Execution validate-configurations of goal org.openmrs.maven.plugins:openmrs-packager-maven-plugin:1.7.0:validate-configurations failed: Plugin org.openmrs.maven.plugins:openmrs-packager-maven-plugin:1.7.0 or one of its dependencies could not be resolved: Could not find artifact org.openmrs.module:initializer-validator:jar:2.5.2 in openmrs-repo (https://openmrs.jfrog.io/artifactory/public)
```

There are two entries in the POM, both labelled `openmrs-repo`, one in repositories and one in pluginRepositories. We are able to find Initializer when included as a dependency (it is a dependency of the `distro` sub-project and this works), but we fail to find it when it is a dependency of a Maven plugin. I'm guessing this is because we have a separate pluginRepositories element, but with no reference to the Mekom repos there, which is why we're getting the unresolved thing. This has nothing to do with repo order or proxying or anything like this. We've just told Maven to only search two repositories for plugins and dependencies, and neither of them have Initializer in them.

Adding the Mekom Nexus references to the pluginRepositories element resolves the issue, at least for me.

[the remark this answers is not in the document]
rbuisson: What's weird here is that when running the build in Gitpod, it really works fine.

It's able to fetch the `initializer-validator 2.5.2` as it should (even with the Mekom repos being under the OpenMRS ones, as well as with using `https://mavenrepo.openmrs.org/releases` instead of `https://openmrs.jfrog.io/artifactory/releases`)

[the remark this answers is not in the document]
ibacher: I might guess that your Gitpod instance had already separately resolved initializer-validator 2.5.2. If the dependency (especially a released dependency) is in the local Maven cache, I don't think Maven worries too much about where it was resolved from.

rbuisson:
Thanks for your last fix @ibacher. Seems to fix the CI issue, are we good to merge now?

ibacher (approved):

ibacher:
I think so, yeah

PR #719, opened 2023-04-27
Release 3.0.0-beta.6

rbuisson:


distro/pom.xml
```diff
@@ -28,15 +28,15 @@
     <initializer.version>2.5.0</initializer.version>
     <webservices.rest.version>2.39.0</webservices.rest.version>
     <addresshierarchy.version>2.15.1</addresshierarchy.version>
-    <idgen.version>4.10.0-SNAPSHOT</idgen.version>
+    <idgen.version>4.10.0</idgen.version>
     <legacyui.version>1.15.0</legacyui.version>
     <metadatamapping.version>1.6.0</metadatamapping.version>
     <metadatasharing.version>1.9.0</metadatasharing.version>
-    <openconceptlab.version>2.1.0-SNAPSHOT</openconceptlab.version>
+    <openconceptlab.version>2.0.0</openconceptlab.version>
```
Ruhanga, line 35: @rbuisson, the 2.1.0 release was given a green light [on talk](https://talk.openmrs.org/t/o3-release-list-of-backend-modules/39494/3). Could we use 2.1.0 instead?

Ruhanga (approved):

rbuisson:
I've just added the revert commit.
@ibacher , is that now suitable for merge?

ibacher:
@rbuisson Could you adjust the revert commit to not roll-back to SNAPSHOT versions of the Maven dependencies? I think we're fine to keep those pinned to release versions and then update to SNAPSHOTs if features are added that we need (that's been our usual pattern). Otherwise, LGTM.

rbuisson:
There it is.

PR #721, opened 2023-04-27
O3-2078: Update to Procedures: ECG,CXR,ChestCT,ChestU/S

gracepotma:


ibacher (approved):

ibacher:
Should we also update the form in the iniz config?

gracepotma:
Yes that's inprogress; I triggered a bug / bad state in the Form Builder by adding 2 concepts to the JSON schema that didn't exist in the distro and got the form stuck in both an infinite loading state and it removed the JSON we needed. Logging that and fixing the form then will do a PR.

PR #726, opened 2023-06-02
Release beta.9

Ruhanga:
This release is composed of the latest Frontend artefacts, _Laboratory Tests_ form fix and maintains the backend artefacts' versions, similar to the previous release's (beta.8) backend versions.


frontend/Dockerfile
```diff
@@ -1,7 +1,7 @@
 # syntax=docker/dockerfile:1.3
 FROM --platform=$BUILDPLATFORM node:18-alpine as dev
 
-ARG APP_SHELL_VERSION=next
+ARG APP_SHELL_VERSION=4.4.1
```
ibacher: This should be 4.5.0, like the rest of the esm-core files

Ruhanga: Thanks for the catch. This should look good now... :-)

PR #735, opened 2023-07-14
Add the docker compose file used for creating pre-filled images

Piumal1999:
## Purpose

The purpose of this pull request is to create a docker-compose file that can be used to create pre-filled Docker images. The newly added docker-compose file enables the creation of a backend environment without using volumes. With this, we can write a Bamboo job that runs the backend environment, waits for data to be generated, and obtains a Docker image that includes the data.

I have named the file docker-compose-no-volumes.yml. Let me know if there are any suggestions for the file name.

Related talk thread: https://talk.openmrs.org/t/using-pre-filled-docker-images-for-running-e2e-tests/40003

### DIFF
--- a/docker-compose-no-volumes.yml
+++ b/docker-compose-no-volumes.yml
@@ -1,0 +1,31 @@
+# This docker compose file is used to create a backend environment without volumes.
+version: "3.7"
+
+services:
+  backend:
+    image: openmrs/openmrs-reference-application-3-backend:${TAG:-nightly}
+    depends_on:
+      - db
+    environment:
+      OMRS_CONFIG_MODULE_WEB_ADMIN: "true"
+      OMRS_CONFIG_AUTO_UPDATE_DATABASE: "true"
+      OMRS_CONFIG_CREATE_TABLES: "true"
+      OMRS_CONFIG_CONNECTION_SERVER: db
+      OMRS_CONFIG_CONNECTION_DATABASE: openmrs
+      OMRS_CONFIG_CONNECTION_USERNAME: ${OPENMRS_DB_USER:-openmrs}
+      OMRS_CONFIG_CONNECTION_PASSWORD: ${OPENMRS_DB_PASSWORD:-openmrs}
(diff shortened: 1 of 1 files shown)


Piumal1999:
Hi @ibacher, Could you please review and merge this?

docker-compose-no-volumes.yml
```diff
@@ -0,0 +1,31 @@
...
+      OMRS_CONFIG_CREATE_TABLES: "true"
+      OMRS_CONFIG_CONNECTION_SERVER: db
+      OMRS_CONFIG_CONNECTION_DATABASE: openmrs
+      OMRS_CONFIG_CONNECTION_USERNAME: ${OPENMRS_DB_USER:-openmrs}
+      OMRS_CONFIG_CONNECTION_PASSWORD: ${OPENMRS_DB_PASSWORD:-openmrs}
+    healthcheck:
+      test: ["CMD", "curl", "-f", "http://localhost:8080/openmrs"]
+      timeout: 5s
+    ports:
+      - 9000:8080
```
ibacher, line 21: Is there a reason not to use 8080 as a default here?

Piumal1999: We used 8080 port to run the frontend (with `yarn start --sources 'packages/esm-*-app/` ). So the local backend was run on a different port.

ibacher: Right... I mean, for this use case, we don't even need an exposed port do we?

Piumal1999: But to check whether the initial data generation is completed, we need an exposed port. Isn't it? 

 (In order to use something similar to `$(curl -s -o /dev/null -w ''%{http_code}'' http://localhost:8080/openmrs/login.htm)" != "200"`)

Piumal1999: I changed the exposed port number to 8080 with [f572b1a]

ibacher (approved):

PR #736, opened 2023-07-19
Release `3.0.0-beta.10`

vasharma05:
Initialize the release for beta.10 version.


distro/pom.xml
```diff
@@ -36,7 +36,7 @@
     <!-- this is request for the OCL UI -->
     <owa.version>1.14.0</owa.version>
     <attachments.version>3.0.0</attachments.version>
-    <referencedemodata.version>2.4.0-SNAPSHOT</referencedemodata.version>
+    <referencedemodata.version>2.4.0</referencedemodata.version>
```
Ruhanga, line 39: Hi @vasharma05, lets use 2.3.1 patch release which I've just triggered.

Ruhanga (changes requested):

Ruhanga (commented):
@vasharma05, lets also change the commit message to `(release) Create release commit for beta.10`

You will also add another release to revert the previous one. Here is [an example](https://github.com/openmrs/openmrs-distro-referenceapplication/pull/726).

distro/pom.xml
vasharma05: 2.3.1, right? Or is it something with patch  in the version as well?

Ruhanga: There is a bug fix coming with version 2.3.1 of the Patient Flags module.

vasharma05:
Hi @Ruhanga !
Please re-review the PR.
Thanks!

Ruhanga (commented):
@vasharma05, could you finally run the  `docker compose build` command on the release commit, inside the root folder, just to make sure the build of the docker images (especially the frontend one) goes through okay/is green?

ibacher:
We also have an updated version of the FHIR2 module to include in this release.

ibacher:
1.10.0 IIRC

rbuisson:
@ibacher  it seems that FHIR2 is right now pointing to `1.9.0` and I guess that's what was used recently. I'm guessing  `1.10.0` was not tested specifically with Ref App.
Do you think we should still update this dependency, right now?

ibacher:
> Do you think we should still update this dependency, right now?

Yes. 1.10.0 is completely backwards compatible but has some important bugfixes that affect runtime resource usage.

ibacher:
Hmmm... skip FHIR2. I need to do some work on it.

vasharma05:
So, can we move ahead with release?

ibacher:
The tag needs to be created and the release published via GitHub to trigger the release.

ibacher (approved):

PR #750, opened 2023-09-29
(chore) O3-2460: Upgrade appointments and queues to latest snapshots and add teleconsultation module

mseaton:
…eleconsultation


distro/pom.xml
```diff
@@ -37,8 +37,9 @@
     <owa.version>1.14.0</owa.version>
     <attachments.version>3.0.0</attachments.version>
     <referencedemodata.version>2.4.0-SNAPSHOT</referencedemodata.version>
-    <queue.version>1.0.0-alpha.1</queue.version>
-    <appointments.version>1.6.0</appointments.version>
+    <queue.version>1.0.0-SNAPSHOT</queue.version>
+    <appointments.version>2.0.0-SNAPSHOT</appointments.version>
+    <teleconsultation.version>2.0.0-SNAPSHOT</teleconsultation.version>
```
ibacher, line 42: Is this unavoidable? Ugh...

ibacher (approved):

distro/pom.xml
mseaton: Hopefully temporary, there is some discussion with the Bahmni team to try to transition this to an aware_of dependency.  If you have opinions about that would be good to weigh in.

PR #754, opened 2023-10-25
(test) O3-2525: Add new GH workflow to run E2E on release PRs

jayasanka-sack:
Ticket: https://issues.openmrs.org/browse/O3-2525

### 🎃 Test PR: https://github.com/jayasanka-sack/openmrs-distro-referenceapplication/pull/4

This GitHub Actions workflow, named "Run E2E Tests on Release PRs," serves as Quality Gate `#4` in the [O3 Release QA pipeline](https://docs.google.com/presentation/d/1k3DH74Mz1Afnrgy2MpwR5HQK5vMpVx0pTfN1na62lvI/edit#slide=id.g165af5ac0be_0_24).

Screenshot 2023-10-24 at 18 46 49

The workflow is designed to automate the end-to-end (E2E) testing process for release pull requests (PRs)
that opened in the format described in here: [How to Release the O3 RefApp](https://wiki.openmrs.org/display/projects/How+to+Release+the+O3+RefApp)

The workflow is conditional and **only runs if the pull request title starts with "(release)"**.

Below is an overview of the key components of the workflow:
Screenshot 2023-10-24 at 18 13 19

## Job: build

This job prepares the test environment, extracts version numbers, builds and runs Docker containers, and uploads
artifacts for use in subsequent E2E testing jobs.

### Checking out to the release commit

The reason for the step:

```yaml
- name: Checkout to the release commit
  run: |
  release_commit_sha=$(git log --oneline | awk 'NR==3 {print $1}')
  git checkout $release_commit_sha
```

is to ensure that the workflow checks out to the specific release commit associated with the pull request. This is
necessary because the pull request contain both release commits and revert commits, and the goal is to specifically
target the release commit for further processing.

Here's an explanation of what this step does:

1. It uses `git log --oneline` to retrieve a list of recent commits in the repository.
2. The `awk 'NR==3 {print $1}'` part is used to extract the SHA hash of the third most recent commit, which is the
   release commit we want to target. This is done by specifying `NR==3`, which means selecting the third line (commit)
   in the log.
   The reason to use the 3rd commit instead of 2nd is that a GitHub Action workflow operates within a context where
   there's an additional merge commit on the head of the pull request.
3. Finally, the workflow checks out to this release commit by using git checkout $release_commit_sha.

By checking out to the specific release commit, it ensures that the subsequent steps of the workflow are based on the
state of the code associated with the release, and not on the revert commit present in the pull request.

## End-to-End Test Jobs

The workflow includes several end-to-end test jobs, each corresponding to a specific components of O3 (frontend
mono-repos). These jobs are structured similarly and are listed below:

* `run-patient-management-e2e-tests`
* `run-patient-chart-e2e-tests`
* `run-form-builder-e2e-tests`
* `run-esm-core-e2e-tests`
* `run-cohort-builder-e2e-tests`

In each "End-to-End Test Job," the workflow first checks out the repository associated with a specific OpenMRS
component. It then downloads Docker images from a previous "build" job, loads these images, and starts an OpenMRS instance.

### Why Check Out to the Tags?

The workflow checks out a specific tagged version of the mono-repo, the tag is imported from the previous "build" job. This is necessary because the goal is to perform end-to-end tests on the codebase that corresponds to a
particular release version, rather than the code at the head of the repository.

Screenshot 2023-10-24 at 17 24 38

### DIFF
--- a/.github/workflows/e2e-on-release.yml
+++ b/.github/workflows/e2e-on-release.yml
@@ -21,9 +21,7 @@ jobs:
           fetch-depth: 0
 
       - name: Checkout to the release commit
-        run: |
-          release_commit_sha=$(git log --oneline | awk 'NR==3 {print $1}')
-          git checkout $release_commit_sha
+        run: git checkout $(git log --grep="^(release)" --format="%H" | head -1)
 
       - name: Extract version numbers from the spa-build-config.json file
         id: versions
@@ -33,7 +31,7 @@ jobs:
         run: docker-compose -f e2e_test_support_files/docker-compose-build.yml up -d
 
       - name: Wait for the backend to start
-        run: while [[ "$(curl -s -o /dev/null -w ''%{http_code}'' http://localhost:8080/openmrs/login.htm)" != "200" ]]; do sleep 10; done
+        run: while [[ "$(curl -s -o /dev/null -w ''%{http_code}'' http://localhost/openmrs/login.htm)" != "200" ]]; do sleep 10; done
 
       - name: Commit and export Containers
         run: sh e2e_test_support_files/commit_and_export_images.sh
@@ -64,7 +62,7 @@ jobs:
       - name: Load Docker Images
         run: |
           gzip -d ${{ steps.tempdir.outputs.tmpdir }}/e2e_release_env_images.tar.gz
-          docker load --input e2e_release_env_images.tar
+          docker load --input ${{ steps.tempdir.outputs.tmpdir }}/e2e_release_env_images.tar
           docker image ls -a
 
       - name: Spin up an OpenMRS Instance
@@ -124,8 +122,8 @@ jobs:
 
       - name: Load Docker Images
         run: |
-          tar xzvf /tmp/e2e_release_env_images.tar.gz
-          docker load --input e2e_release_env_images.tar
+          gzip -d ${{ steps.tempdir.outputs.tmpdir }}/e2e_release_env_images.tar.gz
+          docker load --input ${{ steps.tempdir.outputs.tmpdir }}/e2e_release_env_images.tar
           docker image ls -a
 
       - name: Spin up an OpenMRS Instance
@@ -173,6 +171,10 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
+      - name: Create Temporary Directory to Download Docker Images
+        id: tempdir
+        run: echo "tmpdir=$(mktemp -d)" >> "$GITHUB_OUTPUT"
+
       - name: Download Docker Images
         uses: actions/download-artifact@v3
         with:
@@ -181,8 +183,8 @@ jobs:
 
       - name: Load Docker Images
         run: |
-          tar xzvf /tmp/e2e_release_env_images.tar.gz
-          docker load --input e2e_release_env_images.tar
+          gzip -d ${{ steps.tempdir.outputs.tmpdir }}/e2e_release_env_images.tar.gz
+          docker load --input ${{ steps.tempdir.outputs.tmpdir }}/e2e_release_env_images.tar
           docker image ls -a
 
       - name: Spin up an OpenMRS Instance
@@ -230,6 +232,10 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
+      - name: Create Temporary Directory to Download Docker Images
+        id: tempdir
+        run: echo "tmpdir=$(mktemp -d)" >> "$GITHUB_OUTPUT"
+
       - name: Download Docker Images
         uses: actions/download-artifact@v3
         with:
@@ -238,8 +244,8 @@ jobs:
 
       - name: Load Docker Images
         run: |
-          tar xzvf /tmp/e2e_release_env_images.tar.gz
-          docker load --input e2e_release_env_images.tar
+          gzip -d ${{ steps.tempdir.outputs.tmpdir }}/e2e_release_env_images.tar.gz
+          docker load --input ${{ steps.tempdir.outputs.tmpdir }}/e2e_release_env_images.tar
           docker image ls -a
 
       - name: Spin up an OpenMRS Instance
@@ -287,6 +293,10 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
+      - name: Create Temporary Directory to Download Docker Images
+        id: tempdir
+        run: echo "tmpdir=$(mktemp -d)" >> "$GITHUB_OUTPUT"
+
       - name: Download Docker Images
         uses: actions/download-artifact@v3
         with:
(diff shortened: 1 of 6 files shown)


.github/workflows/e2e-on-release.yml
```diff
@@ -0,0 +1,331 @@
...
+      patient_management_version: ${{steps.versions.outputs.patient_management}}
+      patient_chart_version: ${{steps.versions.outputs.patient_chart}}
+      esm_core_version: ${{steps.versions.outputs.esm_core}}
+      form_builder_version: ${{steps.versions.outputs.form_builder}}
+      cohort_builder_version: ${{steps.versions.outputs.cohort_builder}}
+
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
```
ibacher, line 21:
```suggestion
          fetch-depth: 2
```

.github/workflows/e2e-on-release.yml
```diff
@@ -0,0 +1,331 @@
...
+
+      - name: Checkout to the release commit
+        run: |
+          release_commit_sha=$(git log --oneline | awk 'NR==3 {print $1}')
+          git checkout $release_commit_sha
```
ibacher, line 26: Personally I think the equivalent:

```
git checkout HEAD~2
```

Is a lot clearer...

e2e_test_support_files/extract_tag_numbers.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+
+# Define a function to extract and output a dependency value
+get_repository_tag() {
+  local file="$1"
+  local repo_name="$2"
+  local app="$3"
+  local value
+  value=$(grep "\"$app\":" "$file" | awk -F'"' '{print $4}')
```
ibacher, line 9: You could actually do this just with awk (slightly awkward because we need to escape the `/` in the app name):

```
value=$(awk -F'"' "/${app/\//\/}/ {print \$4}" "$file")
```

.github/workflows/e2e-on-release.yml
```diff
@@ -0,0 +1,331 @@
...
+
+      - name: Commit and export Containers
+        run: sh e2e_test_support_files/commit_and_export_images.sh
+
+      - name: Upload artifact
+        uses: actions/upload-artifact@v3
+        with:
+          name: e2e_release_env_images
+          path: e2e_release_env_images.tar.gz
+          retention-days: 1
```
ibacher, line 46: I'd probably add a "finally" job that just removes the artifact. That's better hygiene, I think.

.github/workflows/e2e-on-release.yml
```diff
@@ -0,0 +1,331 @@
...
+    runs-on: ubuntu-latest
+    needs: build
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Download Docker Images
+        uses: actions/download-artifact@v3
+        with:
+          name: e2e_release_env_images
+          path: /tmp
```
ibacher, line 58: It might be slightly messy, but it's a little better to have a step that uses `mktemp -d` and saves the output in a output variable, which is then used in-place of `/tmp`. This makes the job less dependent on GH Actions providing us with a clean environment.

e2e_test_support_files/docker-compose-build.yml
```diff
@@ -0,0 +1,61 @@
...
+      OMRS_CONFIG_CREATE_TABLES: "true"
+      OMRS_CONFIG_CONNECTION_SERVER: db
+      OMRS_CONFIG_CONNECTION_DATABASE: openmrs
+      OMRS_CONFIG_CONNECTION_USERNAME: ${OPENMRS_DB_USER:-openmrs}
+      OMRS_CONFIG_CONNECTION_PASSWORD: ${OPENMRS_DB_PASSWORD:-openmrs}
+    healthcheck:
+      test: [ "CMD", "curl", "-f", "http://localhost:8080/openmrs" ]
+      timeout: 5s
+    ports:
+      - 8080:8080
```
ibacher, line 49: I don't love directly exposing the backend. If we have the gateway, we should use it. The same curl commands will still work in the same way.

e2e_test_support_files/docker-compose.yml
```diff
@@ -0,0 +1,26 @@
...
+
+services:
+  gateway:
+    image: gateway
+    restart: "unless-stopped"
+    depends_on:
+      - frontend
+      - backend
+    ports:
+      - "8080:80"
```
ibacher, line 11: Is there a reason this is exposed on port 8080 rather than 80?

e2e_test_support_files/commit_and_export_images.sh
```diff
@@ -0,0 +1,16 @@
...
+
+docker commit $frontend_container_id frontend
+docker commit $gateway_container_id gateway
+docker commit $backend_container_id backend
+docker commit $db_container_id db
+
+docker save frontend gateway backend db > e2e_release_env_images.tar
+
+# compress the file (to decrease the upload file size)
+tar czvf e2e_release_env_images.tar.gz e2e_release_env_images.tar
```
ibacher, line 16: Isn't it better to just use `gzip` and `gunzip`? This adds an extra TAR containing a TAR and then compresses that... I mean, it works, but you already have a TAR.

ibacher (commented):
Thanks @jayasanka-sack! This is excellent! A few minor nits, but this LGTM.

jayasanka-sack:
Thank you for your review, @ibacher . The issue with using HEAD~2 is that it actually points to the second most recent commit on the main branch. This is the reason I included `depth:0`, as without it, the PR branch's commits wouldn't be fetched. I'd appreciate your thoughts on this matter.

This is how it's represented in the GitHub workflow context. 

Screenshot 2023-10-25 at 11 11 00

e2e_test_support_files/docker-compose-build.yml
jayasanka-sack: My bad. I used it for testing. I'll remove it.

e2e_test_support_files/docker-compose.yml
jayasanka-sack: I just reused the `.env.example` file. which uses the `8080` port.

ibacher:
@jayasanka-sack Alright... How about something a little cleverer `HEAD^{/\(release)}`

jayasanka-sack:
`git checkout HEAD^{/\(release\)}` checkouts to whatever commit that contains `release`. Any idea?

Btw, how about `git checkout $(git log --grep="^(release)" --format="%H" | head -1)`?

e2e_test_support_files/commit_and_export_images.sh
jayasanka-sack: fixed with 29efd49

.github/workflows/e2e-on-release.yml
jayasanka-sack: fixed with 29efd49

.github/workflows/e2e-on-release.yml
jayasanka-sack: fixed with 29efd49

jayasanka-sack: Deleting artifact seems tricky. While there are marketplace options available for the same, they are used by only a few users.

ibacher:
>  checkouts to whatever commit that contains release. Any idea?

Right... specifically it should checkout the nearest commit to HEAD that contains `(release)`. You could also add the `^` like in the grep to force it be at the start. You may need to quote the

jayasanka-sack:
That worked. Thanks, @ibacher ! Could you please review the PR again?

ibacher (approved):

PR #758, opened 2023-11-17
Rename the patient list app and add new patient lists workspace app

denniskigen:
See https://github.com/openmrs/openmrs-esm-patient-management/pull/884.


ibacher (approved):
We should only merge this after the patient-management one

PR #762, opened 2023-11-21
Release 3.0.0-beta.15

ibacher:


distro/pom.xml
```diff
@@ -36,11 +36,11 @@
...
     <attachments.version>3.0.0</attachments.version>
-    <referencedemodata.version>2.3.1</referencedemodata.version>
-    <queue.version>2.0.0-SNAPSHOT</queue.version>
-    <appointments.version>2.0.0-SNAPSHOT</appointments.version>
-    <teleconsultation.version>2.0.0-SNAPSHOT</teleconsultation.version>
-    <cohort.version>3.6.0-SNAPSHOT</cohort.version>
+    <referencedemodata.version>2.4.0</referencedemodata.version>
+    <queue.version>2.0.0</queue.version>
+    <appointments.version>2.0.0-20231101.130425-7</appointments.version>
+    <teleconsultation.version>2.0.0-20230831.113926-1</teleconsultation.version>
```
vasharma05, line 42: Hi @ibacher ,
Shouldn't we have reverted the SNAPSHOT versions of the backend modules above in the revert release?
Thanks!

ibacher: My policy with backend modules is I only revert the versions once there's a change we want to adopt.

PR #771, opened 2023-12-19
(fix) O3-2605: Update the field type of the Insurance visit attribute

vasharma05:
https://openmrs.atlassian.net/browse/O3-2605


jayasanka-sack (approved):
Thanks @vasharma05 !

PR #772, opened 2023-12-19
O3-2595: Fix Patient Data collection metadata import errors

gracepotma:
Action taken: Manually added UUID to the Patient ID Photo concept Name UUID field, and then to Description UUID field. Then subscription worked. Required multiple tries hence we are going from version 2 of this collection to version 4.

Before: Seeing this error: "Cannot import concept 736e8771-e501-4615-bfa7-570c03f4bef5, tried: Retrying import of concept 736e8771-e501-4615-bfa7-570c03f4bef5 after failure due to 'could not execute statement' [CAUSE]: Column 'uuid' cannot be null"

After: No errors.


jayasanka-sack (approved):
Thanks!

PR #774, opened 2023-12-20
O3-2597: Remove irrelevant mappings causing import errors

gracepotma:
Fix for O3-2597: Removed the 4 mappings to CIEL 164844 ("HIV status of source person") as this was causing the error messages for this collection, and I can find no reason these mappings are needed in this specific collection.
Jira issue: https://openmrs.atlassian.net/browse/O3-2597


jayasanka-sack (approved):
Thanks @gracepotma !

PR #777, opened 2024-01-05
(feat) Allow image to use different frontend setup by modifying spa-build-config.json

ibacher:
This PR attempts to add a feature to make the frontend image re-usable outside of the RefApp context without interfering with the RefApp image. The main idea here is that the spa-build-config.json file is relatively well-understood mechanism to describe a frontend, so if the user supplies a new spa-build-config.json, we simply process it and replace the frontend with that. This is checked for by using the SHA-512 sum of the spa-build-config.json.

I'm less than happy with the story around how this checksum is persisted; specifically, the "base" checksum is part of the image (to prevent needing to rebuild if you're not modifying the spa-build-config.json), but if the spa-build-config.json is modified, although the checksum file is updated, this may not be persisted in the same way.

I'm also unsure that this method of adding to the image is suitable for most use-cases of modifying the frontend.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
 # syntax=docker/dockerfile:1
 
 ### Dev Stage
-FROM openmrs/openmrs-core:dev as dev
+FROM --platform=$BUILDPLATFORM openmrs/openmrs-core:dev as dev
 WORKDIR /openmrs_distro
 
 ARG MVN_ARGS_SETTINGS="-s /usr/share/maven/ref/settings-docker.xml -U"
@@ -25,7 +25,7 @@ RUN mvn $MVN_ARGS_SETTINGS clean
 
 ### Run Stage
 # Replace 'nightly' with the exact version of openmrs-core built for production (if available)
-FROM openmrs/openmrs-core:nightly
+FROM --platform=$BUILDPLATFORM openmrs/openmrs-core:nightly
 
 # Do not copy the war if using the correct openmrs-core image version
 COPY --from=dev /openmrs/distribution/openmrs_core/openmrs.war /openmrs/distribution/openmrs_core/
--- a/frontend/Dockerfile
+++ b/frontend/Dockerfile
@@ -1,7 +1,8 @@
 # syntax=docker/dockerfile:1.3
-FROM --platform=$BUILDPLATFORM node:18-alpine as dev
-
+ARG NODE_VERSION=18
 ARG APP_SHELL_VERSION=next
+
+FROM --platform=$BUILDPLATFORM node:${NODE_VERSION}-alpine as dev
 
 RUN mkdir -p /app
 WORKDIR /app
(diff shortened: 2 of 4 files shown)


ibacher:
@rbuisson So the solution here is actually adding a process to the container with the web server to rebuild things, so there isn't a Docker process involved. Essentially it overwrites the frontend with the new configuration, so it would require a few Docker volumes to make things persist correctly, but there's no additional CI/CD process required.

> Does this process require internet access?

Currently, yes, but only when the shell needs to be rebuilt (so when the s`spa-build-config.json` file changes). Otherwise no internet access is required. There is a way to do this without internet access at all, but it involves a `spa-build-config.json` file that looks something like:

```json
{
  "frontendModules": {
     "@openmrs/esm-login-app": "file:///openmrs/frontend_modules/openmrs-esm-login-app-1.2.3.tgz",
     "@openmrs/esm-primary-navigation-app": "file:///openmrs/frontend_modules/openmrs-esm-primary-navigation-app-1.2.3.tgz"
  }
}
```

You then have to mount each of those `tgz` files in the `/openmrs/frontend_modules` directory. (Fully supporting no network access also means a slight adjustment so that the `openmrs` CLI tool is installed, but that's trivial to do.

> So possibly, does one need to run the frontend once first (during a CI/CD process) to get the new frontend built and commit/publish that new image, or can it be safely done when deploying the application for actual use?

I was trying to avoid the need to commit or publish an image. Doing that is much cleaner, e.g.,

```dockerfile
# syntax=docker/dockerfile:1.3
FROM --platform=$BUILDPLATFORM node:18-alpine as dev

ARG APP_SHELL_VERSION=next
ARG FRONTEND_VERSION=nightly

RUN mkdir -p /app
WORKDIR /app

COPY spa-build-config.json .

ARG CACHE_BUST
RUN npx --legacy-peer-deps openmrs@${APP_SHELL_VERSION:-next} assemble --manifest --mode config --config spa-build-config.json --target ./spa
RUN npx --legacy-peer-deps openmrs@${APP_SHELL_VERSION:-next} build --build-config spa-build-config.json --target ./spa
RUN if [ ! -f ./spa/index.html ]; then echo 'Build failed. Please check the logs above for details. This may have happened because of an update to a library that OpenMRS depends on.'; exit 1; fi

FROM --platform=$BUILDPLATFORM openmrs/openmrs-reference-application-3-frontend:${FRONTEND_VERSION}

# clear the default files installed by nginx
RUN rm -rf /usr/share/nginx/html/*

COPY --from=dev /app/spa /usr/share/nginx/html
```

But obviously this then requires a container registry to store it and it needs to be updated at appropriate intervals, etc.

The idea here was for this to be doable on an already built image. There's no need to run the frontend once before hand. It just requires the following volumes:

```yaml
volumes:
  - path/to/spa-build-config.json:/openmrs/spa-build-config.json
  - <volume>:/openmrs/span-build-config.json.sha512 # i don't actually know if this is possible; this might need to be a bind mount
  - <volume>:/usr/share/nginx/html
```

rbuisson:
From yesterday's discussion, we might try rather to:
- make the `spa-build-config.json` a published artifact (Maven likely)
- make the `assemble` command able to use an override file which would specify includes + excludes.

That way, groups wanting to override the Ref App list of ESMs would be able to depend on the published `spa-build-config.json` and provide their own overriding config, reassemble and build the frontend (likely via CI/CD), then mount that to the O3 community Docker image.

@ibacher  is the above correctly reflects what we said?

ibacher:
@rbuisson Yes. It turns out this will be slightly more complicated than I originally anticipated, because we were doing some funky, but inconsequential things in the assemble command.

PR #779, opened 2024-01-22
O3-2753: Separate spa-build-config from spa-assemble-config

ibacher:
See https://openmrs.atlassian.net/browse/O3-2753

This separates out the assemble configuration from the build configuration. Strictly speaking, this isn't a required change, but since this repo is something of an example of how to build O3 and we're going to start publishing artefacts that looks more like this, it seemed good to add the setup to do this.


rbuisson (approved):
LGTM.
Do you think it could make sense to document the `excludedFrontendModules`feature by providing an empty array for instance?
Anyway, looks good as it is.

PR #780, opened 2024-01-24
O3-2753: Add ability to publish frontend configuration to Maven

ibacher:
Apparently I lied and there was more. This adds a POM and some structure to allow us to publish the `spa-assemble-config.json` (with version numbers) and the `spa-build-config.json` files as a Maven artifact.

In CI we will have something like this:

```sh
docker create --name=openmrs-frontend-container openmrs/openmrs-reference-application-3-frontend:nightly
docker cp openmrs-frontend-container:/usr/share/nginx/html/spa-assemble-config.json src/main/resources/spa-assemble-config.json
docker rm openmrs-frontend-container
```

Note that this means that the POM in the `frontend` directory is intended only as a "publication POM" and not a "actual build".

### DIFF
--- a/frontend/Dockerfile
+++ b/frontend/Dockerfile
@@ -1,4 +1,7 @@
 # syntax=docker/dockerfile:1.3
+#--------------------------------------------
+# Dev Stage - Assembles and Builds Frontend
+#--------------------------------------------
 FROM --platform=$BUILDPLATFORM node:18-alpine as dev
 
 ARG APP_SHELL_VERSION=next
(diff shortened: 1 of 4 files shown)


rbuisson (commented):
@ibacher , `mvn clean package -Pfrontend` does not seem to output a file. There is a Zip but it looks like it's empty.

Is that supposed to be?

ibacher:
@rbuisson That's kind of what I meant that this is a "publication POM" not a "actual build", that is the POM here depends on the Docker build rather than the other way around. The main build for the RefApp is still `docker compose build`, there's just an additional step that will extract the necessary file from the built container.

ibacher:
@rbuisson As an update, you can see the published Zip [here](https://openmrs.jfrog.io/ui/repos/tree/ViewSource/snapshots/org/openmrs/distro/referenceapplication-frontend/3.0.0-SNAPSHOT/referenceapplication-frontend-3.0.0-20240126.155827-1.zip/spa-assemble-config.json).

rbuisson:
Thanks. I understood it the other way around. As in "publication POM" was to only spit a file that will be published, and not used anywhere else in the project.

I can see the published artefact 👍🏻 .

PR #787, opened 2024-03-11
(feat) O3-2948 Add frontend config to RefApp

chibongho:
As mentioned in [O3-2948](https://openmrs.atlassian.net/browse/O3-2948), we want to be able to configure for demo data in the frontend. This PR adds the ability to do that.

 Tested as follows: 
- hard-code a different brand color in `config-core_demo.json`
- do a `docker-compose build` then `docker-compose up`. 
- verify that a different brand color shows up when going to `http://localhost`

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -16,7 +16,7 @@ services:
     environment:
       SPA_PATH: /openmrs/spa
       API_URL: /openmrs
-      SPA_CONFIG_URLS:
+      SPA_CONFIG_URLS: config-core_demo.json
       SPA_DEFAULT_LOCALE:
     healthcheck:
       test: ["CMD", "curl", "-f", "http://localhost/"]
--- a/README.md
+++ b/README.md
@@ -56,4 +56,6 @@ To help us keep track of things, we ask that you suffix any files you add with e
 those that should be part of the core package. For example, a form named `test_form.json` would become
 `test_core-core_demo.json`.
 
+Frontend configuration can be found in `frontend/config-core_demo.json`.
+
 Thanks!
--- a/frontend/Dockerfile
+++ b/frontend/Dockerfile
@@ -36,5 +36,6 @@ RUN chmod +x /usr/local/bin/startup.sh
 COPY nginx.conf /etc/nginx/nginx.conf
 
 COPY --from=dev /app/spa /usr/share/nginx/html
+COPY config-core_demo.json /usr/share/nginx/html
 
 CMD ["/usr/local/bin/startup.sh"]
--- a/frontend/config-core_demo.json
+++ b/frontend/config-core_demo.json
@@ -1,0 +1,5 @@
+{
+  "@openmrs/esm-styleguide": {
+    "Brand color #1": "#005d5d"
+  }
+}


mseaton (approved):
LGTM, thanks @chibongho .  Interested in how @ibacher and others see this.

docker-compose.yml
```diff
@@ -16,7 +16,7 @@ services:
     environment:
       SPA_PATH: /openmrs/spa
       API_URL: /openmrs
-      SPA_CONFIG_URLS:
+      SPA_CONFIG_URLS: config-core_demo.json
```
ibacher, line 19:
```suggestion
      SPA_CONFIG_URLS: /openmrs/spa/config.json
```

ibacher: Note I'm also renaming the file. I don't know that we need the same naming scheme here.

ibacher (commented):
Couple of quick things. Need to think about this a bit. I think, in an ideal world, we'd have this as something that is just part of a "standard" frontend build rather than a separate file added to the distro.

docker-compose.yml
mseaton: Isn't this file already relative to /openmrs/spa/ by default @ibacher ?  Can't we just point this to "config.json" then?

chibongho:
@ibacher Do you have new thoughts on this? I'm happy to prototype something if you can suggest a different place to put this.

ibacher:
I actually thought this had been merged in. Might as well get it in for now and we can work on a slightly different approach later.

pgorrindo:
Thanks @chibongho for your work on this feature!

For the life of me, however, I can't get it to work. 

On a fresh clone of the `openmrs-distro-referenceapplication` repo, I've locally edited `frontend/config-core_demo.json` with

```json
"@openmrs/esm-styleguide": {
    "Brand color #1": "#000000",
    "Brand color #2": "#000000",
    "Brand color #3": "#000000"
  }
```

(note the `#000000` is just for trying to get this feature to work, I'd never actually make all the colors black intentionally)

and then several different tries of `docker-compose build` and `docker-compose up` and wiping my Docker containers and volumes for good measure and trying different browsers in case some styling were cached, etc -- but the styleguide-specified colors in `config-core_demo.json` aren't getting picked up. Instead, in the client, I see:

Screenshot 2024-04-21 at 18 30 30

Which is confusing, since the message on the right ("The current value comes from /openmrs/spa/config-core_demo.json") suggests my local config colors ought to be picked up from there -- but then clearly, they aren't (i.e., would expect the color value on the left to be `#000000`)

Any thoughts on what might be going on here?

@mseaton or @ibacher any thoughts from either of you on this, since you previously reviewed this PR?

Apologies if I'm misunderstanding something here. (Also apologies if you all would prefer I posted this at https://talk.openmrs.org -- since you have issues turned off for this repo.)

thanks for any thoughts or insight!

mseaton:
@pgorrindo - I'd have to see your entire setup to understand more.  Are you sure you are actually running the docker container that you have built locally, and not running the default "qa" tag of the docker container?  What do your docker-compose files look like and what command are you running to start the containers up?

chibongho:
@pgorrindo another thing I would try is to do do a hard refresh of the page (Ctrl+Shift+R). Also, you can do "view source" in the browser to view the HTML and see of the config is set correctly. There should be a relatively human-readable `<script>` tag with a `configUrls` defined in a JSON object. Make sure that it looks something like this:
`configUrls: ["/openmrs/spa/config-core_demo.json"]`

PR #789, opened 2024-03-13
(feat) O3-2204: Add and configure demo queues to the refapp

mseaton:
This PR tags the "Outpatient Clinic" as a "Queue Location", adds 2 demo queues (one for each of the configured services of "Triage" and "Consultation") associated with this Location, and configures the necessary Queue global properties to reflect the installed Concepts.  Note:  This requires a SNAPSHOT version of initializer to install queues, it would be good if we can get that released @ibacher, ideally with my change to the OclLoader, but I don't see that as necessary to merge this PR.

With this change, one will be able to use the service queues esm in the refapp demo / dev3.


mseaton:
Ping on this? @gracepotma / @ibacher ?

chibongho (approved):
LGTM

gracepotma:
What would this mean for implementations who do not use Iniz? Do they end up in a weird state?

distro/configuration/locations/locations-core_demo.csv
gracepotma, line 1: Sorry what changed here?

distro/configuration/globalproperties/globalproperties-core_demo.xml
```diff
@@ -5,4 +5,30 @@
...
+    <!-- https://app.openconceptlab.org/#/orgs/openmrs/sources/DemoQueueConcepts/concepts/DemoQueue-Priority/ -->
+    <globalProperty>
+        <property>queue.priorityConceptSetName</property>
+        <value>78063dec-b6d8-40c1-9483-dd4d3c3ca434</value>
+    </globalProperty>
+
+    <!-- https://app.openconceptlab.org/#/orgs/openmrs/sources/DemoQueueConcepts/concepts/DemoQueue-Status/ -->
+    <globalProperty>
+        <property>queue.statusConceptSetName</property>
+        <value>d60ffa60-fca6-4c60-aea9-a79469ae65c7</value> <!-- PIH Queue Status Set -->
```
ibacher, line 26: Should probably get rid of the comment mentioning PIH here.

distro/configuration/locations/locations-core_demo.csv
ibacher: Added a column

ibacher:
> What would this mean for implementations who do not use Iniz?

It's just global properties and concept sets, so they'd need to load those in some other fashion.

distro/configuration/globalproperties/globalproperties-core_demo.xml
mseaton: Oh, good catch - removed @ibacher

PR #798, opened 2024-03-22
O3-2204 - Bump queue module to latest snapshot to resolve frontend de…

mseaton:
…pendency, and configure priority and status for refapp based on designs.


chibongho (approved):
LGTM. Thanks!

PR #801, opened 2024-04-03
Bundle Client registry

sherrif10:
Ticket  https://openmrs.atlassian.net/browse/HIE-7

 cc @samuelmale @dkayiwa

### DIFF
--- a/distro/distro.properties
+++ b/distro/distro.properties
@@ -28,3 +28,4 @@ omod.serialization.xstream.type=omod
 omod.ordertemplates=${ordertemplates.version}
 omod.patientflags=${patientflags.version}
 omod.o3forms=${o3forms.version}
+omod.clientregistry=${clientregistry.version}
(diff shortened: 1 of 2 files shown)


sherrif10:
Hello @samuelmale . Do we really need to bundle this client registry. i want to say that am following the older process, Not quite sure if its the same proccess since we are running on 03 branch.

tendomart:
wonder why it's complaining of `1.1.8-SNAPSHOT`

tendomart:
@sherrif10  i think this is for the backend support.

tendomart:
I think no version was released to JFROG, that's why it's complaining

sherrif10:
Yes @tendomart .Thats right.

samuelmale:
@sherrif10 unfortunately I don't think we are yet to the point of bundling this module in the refApp

sherrif10:
Thanks, i will close this for now.

PR #804, opened 2024-04-18
O3-3078: Removed and replace Lab Order Entry form with configured Lab Order basket.

Ruhanga:
Ticket: https://openmrs.atlassian.net/browse/O3-3078


mks-d (approved):
Approving this since we have it working on a temporary fork for a client.

PR #805, opened 2024-04-19
(feat) O3-2972: Validate number fields whose concepts don't allow decimal values

denniskigen:
This PR adds validation to the Laboratory Test Results form that ensures fields of type [number](https://ampath-forms.vercel.app/docs/field-types-reference#number) whose backing concepts don't allow decimal values show an error when decimals are typed in. For example, the backing concept for `Platelets` does not allow decimal values:

323766319-4f8adbca-0a73-43ea-9e0a-d77d5b705912

This validation logic was recently introduced to the Angular form engine [here](https://github.com/openmrs/openmrs-ngx-formentry/pull/135).

## Video of the validation in practice

https://github.com/openmrs/openmrs-distro-referenceapplication/assets/8509731/f675af6d-d540-4349-9ac1-e48a4e120a37

This change solves this error shown when submitting the form after filling in values that lie within the related reference ranges for the backing concepts but without taking into account whether the fields support decimal values or not:

323768777-bca7dcfe-a046-482b-97ba-9f37b094def9

## Related issue

https://openmrs.atlassian.net/browse/O3-2972

## Other 

I don't know what the versioning scheme in use here is. I've elected to increment the version by 1. More details about the issue are pointed out in [this PR](https://github.com/openmrs/openmrs-ngx-formentry/pull/135).

### DIFF
--- a/distro/configuration/ampathforms/TestResultsEntryFormv2-core_demo.json
+++ b/distro/configuration/ampathforms/TestResultsEntryFormv2-core_demo.json
@@ -1226,55 +1226,6 @@
               }
             },
             {
-              "label": "Amylase (IU/L)",
-              "type": "obs",
-              "required": false,
-              "id": "manualEntryAmylase",
-              "questionOptions": {
-                "disallowDecimals": true,
-                "rendering": "number",
-                "concept": "1299AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA",
-                "conceptMappings": [
-                  {
-                    "relationship": "SAME-AS",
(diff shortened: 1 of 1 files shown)


denniskigen:
Also, @gracepotma, do you know why the Amylase (IU/L) question is included twice on the Chemistry page?

ibacher:
We can probably get rid of the second copy of the Amylase question. Form versioning is weird. Technically, the requirements are: the first character must be a number and each version should be unique, but since this changes things enough, bumping it to 3 seems called for.

ibacher (approved):

PR #808, opened 2024-04-25
O3-2599: Fix majority of Lab Test metadata errors

gracepotma:
Ticket: https://openmrs.atlassian.net/browse/O3-2599 
This PR updates the "BasicLabTest" collection from version 4.2 to version 4.6. What changed was all related to troubleshooting error messages:
* Updated mappings from OMRS source version (needed to receive 2023 updates to custom concepts from the OpenMRS Source)
* Removed the ICD code for Other faecal abnormalities which was triggering errors. 
* Removed references to chikungunya (they were not expected), this resolved several error messages

This is now down to just 1 error in v 4.6: “Cannot save mapping https://api.openconceptlab.org/orgs/openmrs/sources/ExampleLabFilter/mappings/4660546/ [CAUSE]: Column 'uuid' cannot be null”. It seems I fixed the missing mapping UUID back in December, but I can't get the mapping in the collection to update. I'll continue to follow up with Burke and the OCL team who have been trying really hard to unblock us. 

Otherwise, **a huge special thank you to @mseaton** as I believe his work on the OCL OMOD this year also largely contributed to resolving many of the original error messages I was wrestling with.


mseaton (commented):
Thanks for flagging me @gracepotma - glad to hear things are improving!  Obviously not much to review here between 2 binary files, but it is interesting looking at the filenames.  they show a change from version 4.2 to 4.6 but it's strange that they both end with that date/time from 2023-11-29.  Probably expected, but also worth understanding, especially if you are still seeing behavior you can't explain.

ibacher (approved):

ibacher:
Hmmm... is it no longer possible to get a zip file from OCL? If so, we probably need to adjust the module.

ibacher:
I've released a new version of the concept set and manually confirmed the UUID for 4660546 is in the JSON.

PR #810, opened 2024-05-08
enable showing upcoming appointments on the visit note

mogoodrich:


frontend/config-core_demo.json
```diff
@@ -32,5 +32,8 @@
     },
     "visitQueueNumberAttributeUuid": "",
     "defaultFacilityUrl": ""
+  },
+  "@openmrs/esm-styleguide": {
+    "Brand color #1": "#005d5d"
```
mogoodrich, line 37: OCD, I alphabetized

frontend/config-core_demo.json
```diff
@@ -1,6 +1,6 @@
 {
-  "@openmrs/esm-styleguide": {
-    "Brand color #1": "#005d5d"
+  "@openmrs/esm-patient-chart-app": {
+    "showUpcomingAppointments": true
```
mogoodrich, line 3: enabling this configures the visit note so that it will show upcoming appointments when entering a visit note.  Seemed like it would make sense to turn this on in the refapp for demo purposes?  And this seems like the proper place to configure the refapp?

mseaton (approved):
OK by me, but defer to those who know more about this feature and what we want in the refapp

mogoodrich:
Coincidentally, Usama just asked me if an appointments bug I filed is reproducible on dev3... and it isn't without this change, so I'm going to merge this in, but @gracepotma  @ibacher  feel free to weigh in with objections.  Alternatively, we could just make the default value for this property "true", but that would be a behavior change.

PR #815, opened 2024-06-11
O3-1583: Cause of Death concept with sample answers

gracepotma:
Includes the following:
image

Ticket: https://openmrs.atlassian.net/browse/O3-1583


denniskigen (approved):
LGTM! Thanks!

denniskigen:
It looks like the remaining concern in dev3 is that the `causeOfDeath` global property doesn't exist @ibacher 
 @gracepotma:

CleanShot 2024-06-11 at 10  34 27@2x

gracepotma:
Was causeOfDeath as a Global Property supposed to somehow be introduced by the metadata...? I'm guessing that's a separate issue?

denniskigen:
No, Grace. I was alluding to it for posterity so I can discuss a fix with Ian later.

PR #816, opened 2024-06-18
(release) Release 3.0.0-beta.21

vasharma05:
This PR cuts a new release Beta 21.

This release bumps the following versions:
1. openmrs-core@4.6.6
2. Rest webservices @ 2.44.0


distro/pom.xml
```diff
@@ -24,9 +24,9 @@
     <!-- track module versions
          we do so here so that we can utilise Maven to track updates, etc. -->
     <fhir2.version>2.0.0</fhir2.version>
-    <openmrs.version>2.6.6-SNAPSHOT</openmrs.version>
-    <initializer.version>2.7.0-SNAPSHOT</initializer.version>
-    <webservices.rest.version>2.44.0-SNAPSHOT</webservices.rest.version>
+    <openmrs.version>2.6.6</openmrs.version>
+    <initializer.version>2.7.0-20240521.045407-11</initializer.version>
```
ibacher: Can we get a non-SNAPSHOT release version of Iniz? Relying on SNAPSHOTs is causing us a lot of pain.

jayasanka-sack:
We can safely ignore the following E2E test failures for these reasons:

1. **Patient Management:**
   The test is failing due to a small backend change involving the patient details endpoint. Although the test has already been updated, this job is running outdated tests from the time of the frontend module's release. Therefore, we can ignore this failure.

2. **Patient Chart:**
   The failure is because the tests were still using the angular form engine, whereas this release uses RFE.

3. **Core:**
   There is a minor failure due to an issue with an assertion.

distro/pom.xml
vasharma05: Hi @ibacher!
What will be the version which should be used?

This is the latest Snapshot version after the last commit's release job

vasharma05: image

May I change the initializer version 2.6.0?

jayasanka-sack: I was thinking the same; it would be a mess to rebuild if those were cleaned up.

There are three other snapshots: Queue, Appointments, and Teleconsultation. I wonder why we reached that conclusion in the last release.

gracepotma:
>Can we get a non-SNAPSHOT release version of Iniz? Relying on SNAPSHOTs is causing us a lot of pain.
Agreed @ibacher - but let's do this as a second step. @jayasanka-sack has agreed to follow through on that. 
Rationale: Given we've been using SNAPSHOT release versions (apparently?!) up until this moment, it seems unideal to force the SNAPSHOT versions for this urgent change. But 100% agreed lets make that change; huge thanks to Jayasanka for committing to see this through :)

jayasanka-sack (approved):

gracepotma:
Thank you Jaysanka for explaining those test failures!! <3 Very reassuring.

ibacher:
The concern with Iniz is that about SNAPSHOTs per se. It's that Mekom's Maven instance only keeps a few SNAPSHOTs around, meaning that these releases will break (you can see various messages about people unable to install O3 because the specific SNAPSHOT version we've been relying on no longer exists).

PR #826, opened 2024-07-10
(fix): Updated nginx.conf to allow client upload upto 10MB size objects

njiddasalifu:
I added this to fix the issue: `413 Request Entity Too Large` when trying to install some modules.


njiddasalifu:
@ibacher @robbyoconnor I added this PR just to easily avoid this upload permission for larger objects.

ibacher:
Installing modules via the web UI is not really something we want to encourage. You're free to make this change locally, but I don't think this belongs in the reference version.

PR #832, opened 2024-07-23
O3-3192: Add esm-bed-management-app

brandones:
Follow-up to https://github.com/openmrs/openmrs-esm-patient-management/pull/1230


ibacher:
I presume we need the Bed Management backend module to make this work too?

brandones:
Yes

mseaton (changes requested):
Well, I wouldn't add the esm if we haven't added the backend module, and I don't see it in the distro properties.

PR #862, opened 2024-11-07
(feat) O3-3886: Demo Ward App in the Ref App

mogoodrich:
add bed management ESM
(add new encounter types for Bed Assignment and Cancel ADT)


mogoodrich:
Two new encounter types we will need for features we are adding to the ward app... 1) the ability to swap beds a patient is in (within the same ward) and 2) the ability to cancel a transfer request

chibongho (approved):
Thanks!

mogoodrich:
Merging this in... @chibongho  if during the course of the functionality you are working on we end up changing the modelling, can you remember to change it here as well (or at least ticketing)?  Thanks!

PR #868, opened 2024-11-19
(chore) add visit location tags

mogoodrich:
To be consistent with the EMR-API module convention, every location tagged as a login location must also have a visit location tag (for single-location sites) *or* have a visit location in it's ancestor tree.


mogoodrich:
Thanks @ibacher !

PR #869, opened 2024-11-22
(chore) set restrictByVisitLocationTag parameter to true

mogoodrich:
This will update the Reference Application to use the newly-implemented functionality to restrict visits to those tagged as visit locations (and to default the visit location to the "nearest" visit location associated with the session location), following the O2 Ref App model.

### DIFF
--- a/frontend/config-core_demo.json
+++ b/frontend/config-core_demo.json
@@ -1,5 +1,6 @@
 {
   "@openmrs/esm-patient-chart-app": {
+    "restrictByVisitLocationTag": true,
     "showUpcomingAppointments": true
   },
   "@openmrs/esm-service-queues-app": {


ibacher (approved):

mogoodrich:
Thanks @ibacher , I confirmed that this is place and seems to be working (after a ctrl-reload).

@gracepotma  lmk if you have any questions around this, but, basically, when starting a visit you can now only choose a location that is tagged as a "Visit Location".  In most cases (like with all the "Site x" locations) the locations have been tagged as both Visit and Login locations so you won't see much change.  What is different is for the 5 locations that I've made children of "Ubuntu Hospital" (Inpatient Ward, Outpatient Ward, Wards 1, 2 & 3) you won't be able to choose them as visit locations, and if you are logged into those locations and try to start a visit, the default visit location is set to "Ubuntu Hospital" (representing the fact that your overall visit is at Ubuntu Hospital, though you may have encounters at various different locations within it).

PR #879, opened 2025-01-22
[WIP] OZ-402: Add `config-core_demo.json` in build

VaishSiddharth:


rbuisson:
LGTM.

rbuisson:
Generally, I think the common practice is rather to copy the file as a resource into the build folder as a project build step rather than in the assembly file.
Then the assembly stays generic and copies everything that’s been prepared in the build directory.

ibacher:
So a few things here:

* I'm going to close this because we're going to be moving this into a content package for the up-coming release, so I don't think it's worth anyone's time getting this working correctly.
* If we were to retain this, we would want to have this published as a separate Maven artifact, and not part of the same artifact we're currently publishing.
* As a general rule, we strongly prefer that PRs to OpenMRS repos only reference OpenMRS Jira issues, and not internal issues, so that the reason for the change is recorded in OpenMRS's public history

ibacher:
So a few things here:

* I'm going to close this because we're going to be moving this into a content package for the up-coming release, so I don't think it's worth anyone's time getting this working correctly.
* If we were to retain this, we would want to have this published as a separate Maven artifact, and not part of the same artifact we're currently publishing.
* As a general rule, we strongly prefer that PRs to OpenMRS repos only reference OpenMRS Jira issues, and not internal issues, so that the reason for the change is recorded in OpenMRS's public history

PR #882, opened 2025-01-30
O3-4396: Synced E2E tests with RefApp config

Lemeri123:


Lemeri123:
hello @ibacher my PC keeps disturbing me and I messed up with my commits and everything thats why I closed the other PR and also I ended up making a mistake in this one but anyway its all good now.

capernix (commented):
The change is minor but required.
Before the change the SPA_CONFIG_URLS variable was not set to any specific value
Afterwards it is now set to the json file specified, which is set for the frontend service.
The frontend service will use the specified configuration file during its startup and runtime.
This is a required PR that should be merged.

PR #892, opened 2025-02-18
Remove patient list app

ibacher:


denniskigen (approved):

gracepotma:
Darn - I was too slow in responding. I didn't want this removed in the O3 RefApp. Both Vero and I tested it on test 3 and o3 and it was working perfectly as expected. (See comment on Jira here.)
https://openmrs.atlassian.net/browse/O3-4456?focusedCommentId=146296 

So: Based on that thread, it seems this was broken due to the Platform 2.7 upgrade. 
So I propose that we fix this underlying issue, either by:
1) Keeping this out of the 3.3 release, to meet our deadline, and fix the issue for the 3.4 release (unideal but okay so long as 3.4 is indeed in 1 month, not 3 :P ).
2) Fix the underlying issue before the 3.3 release.

PR #893, opened 2025-02-19
(chore) O3-4460: Prevent over-dispensing medications

denniskigen:
Sets the `restrictTotalQuantityDispensed` property to `true` for the dispensing app. This ensures that the user cannot dispense more medications than the quantity prescribed. The dispense order will be marked as Complete once the quantity dispensed matches the quantity prescribed.

See [this issue thread](https://openmrs.atlassian.net/browse/O3-4460) for more context.

### DIFF
--- a/frontend/config-core_demo.json
+++ b/frontend/config-core_demo.json
@@ -1,4 +1,9 @@
 {
+  "@openmrs/esm-dispensing-app": {
+    "dispenseBehavior": {
+      "restrictTotalQuantityDispensed": true
+    }
+  },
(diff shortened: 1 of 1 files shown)


NethmiRodrigo (approved):
LGTM! Thanks @denniskigen!

denniskigen:
Looks like this leaves the app in an unusable state. We'll revert this change for now and reconsider the approach in the next release cycle.

CleanShot 2025-02-20 at 10  30 23@2x

PR #897, opened 2025-02-22
(chore): Remove unreleased modules

NethmiRodrigo:
This PR removes the following frontend modules from the spa-assemble.config:
- Billing
- Stock management
- Ward 
- Bed management

Should we leave in [the reports app ](https://github.com/openmrs/openmrs-esm-admin-tools/pull/45) that was introduced to the admin tools recently?

### DIFF
--- a/frontend/spa-assemble-config.json
+++ b/frontend/spa-assemble-config.json
@@ -34,8 +34,7 @@
     "@openmrs/esm-primary-navigation-app": "latest",
     "@openmrs/esm-service-queues-app": "latest",
     "@openmrs/esm-system-admin-app": "latest",
-    "@openmrs/esm-user-onboarding-app": "latest",
-    "@openmrs/esm-reports-app": "latest"
+    "@openmrs/esm-user-onboarding-app": "latest"
   },
   "excludedFrontendModules": []
 }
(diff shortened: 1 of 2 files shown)


denniskigen:
I think we can remove that too if we're removing older apps that haven't been fully QA'd.

denniskigen (approved):

frontend/spa-assemble-config.json
```diff
@@ -36,9 +35,6 @@
     "@openmrs/esm-service-queues-app": "latest",
     "@openmrs/esm-system-admin-app": "latest",
     "@openmrs/esm-user-onboarding-app": "latest",
-    "@openmrs/esm-ward-app": "latest",
-    "@openmrs/esm-stock-management-app": "latest",
-    "@openmrs/esm-billing-app": "latest",
```
dkayiwa, line 41: Are we also removing the backend billing module?

NethmiRodrigo: Historically, I don't think we have.

ibacher: We should also do so.

NethmiRodrigo: I think the plan is to get it in the ref app for 3.4.0, because there's some more testing and work left to do

ibacher: Oh... ok.

PR #901, opened 2025-02-25
(fix) Update distro.properties file

NethmiRodrigo:
In #897 I forgot to update the distro.properties file to remove the backend dependencies. This PR fixes it as otherwise it would give an error since the`{module.version}` would be undefined for the modules I removed from the `pom.xml` file


NethmiRodrigo:
I don't know how the builds passed for the previous PR, I got the error when I tried to build it locally

denniskigen (approved):

dkayiwa (approved):

PR #1007, opened 2026-04-23
(fix) add step to build nightly-with-data image

chibongho:
Generated by Claude; haven't tested yet.

### DIFF
--- a/.github/workflows/build-backend.yml
+++ b/.github/workflows/build-backend.yml
@@ -95,7 +95,6 @@ jobs:
     name: Build and Publish the Backend Image Without Demo Content
     if: github.repository_owner == 'openmrs'
     runs-on: ubuntu-latest
-    timeout-minutes: 30
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
 
@@ -151,6 +150,7 @@ jobs:
     name: Build and Publish Nightly With Data Images
     if: github.repository_owner == 'openmrs' && (github.event.inputs.docker_image_tag || 'nightly') == 'nightly'
     runs-on: ubuntu-latest
+    timeout-minutes: 30
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
 
@@ -167,7 +167,7 @@ jobs:
         run: docker compose -f docker-compose-no-volumes.yml up -d
 
       - name: Wait for OpenMRS to be ready
-        run: while [[ "$(curl -s -o /dev/null -w ''%{http_code}'' http://localhost/openmrs/login.htm)" != "200" ]]; do sleep 10; done
+        run: while [[ "$(curl -s -o /dev/null -w ''%{http_code}'' http://localhost/openmrs:8080/login.htm)" != "200" ]]; do sleep 10; done
 
       - name: Commit containers
         run: |
--- a/.github/workflows/build-backend.yml
+++ b/.github/workflows/build-backend.yml
@@ -163,7 +163,7 @@ jobs:
         run: docker compose -f docker-compose-no-volumes.yml pull
 
       - name: Start backend and database
-        run: docker-compose -f docker-compose-no-volumes.yml up -d
+        run: docker compose -f docker-compose-no-volumes.yml up -d
 
       - name: Wait for OpenMRS to be ready
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/build-backend.yml
```diff
@@ -145,3 +145,58 @@ jobs:
...
+        run: docker-compose -f docker-compose-no-volumes.yml pull
+
+      - name: Start backend and database
+        run: docker-compose -f docker-compose-no-volumes.yml up -d
+
+      - name: Wait for OpenMRS to be ready
+        run: |
+          i=300
+          while [[ $i -gt 0 ]]; do
+            status=$(curl -s -o /dev/null -w '%{http_code}' http://localhost:8080/openmrs/login.htm)
```
jayasanka-sack:
```suggestion
            status=$(curl -s -o /dev/null -w '%{http_code}' http://localhost/openmrs/login.htm)
```

the gateway runs on 80.

.github/workflows/build-backend.yml
```diff
@@ -145,3 +145,58 @@ jobs:
...
+          docker commit $db_container_id openmrs/openmrs-reference-application-3-db:nightly-with-data
+
+      - name: Push images
+        run: |
+          docker push openmrs/openmrs-reference-application-3-backend:nightly-with-data
+          docker push openmrs/openmrs-reference-application-3-db:nightly-with-data
+
+      - name: Cleanup
+        if: always()
+        run: docker-compose -f docker-compose-no-volumes.yml down -v
```
jayasanka-sack: I don’t think we need this cleanup step here since it’s the last step of the job and the GitHub runner gets torn down after the job.

This would be useful if we were running on a self hosted runner but I don’t expect that anytime soon.
(or if there were steps after this)

jayasanka-sack (commented):
Thanks @chibongho!

This looks good to me, I left two comments.

.github/workflows/build-backend.yml
jayasanka-sack: Ah my bad, yes the override exposes the backend and it's running on 8080

.github/workflows/build-backend.yml
```diff
@@ -145,3 +145,58 @@ jobs:
           tags: |
             ${{ env.IMAGE }}:${{ env.DOCKER_IMAGE_TAG }}-no-demo
             ${{ env.IMAGE }}:${{ env.DOCKER_IMAGE_TAG_VERSION }}-no-demo
+
+  build-nightly-with-data:
+    name: Build and Publish Nightly With Data Images
+    needs: build-and-publish
```
jayasanka-sack, line 151: This is independent from `build-and-publish` job. So we can get this line removed.
```suggestion
```

.github/workflows/build-backend.yml
```diff
@@ -145,3 +145,57 @@ jobs:
...
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+
+      - name: Log in to Docker Hub
+        uses: docker/login-action@4907a6ddec9925e35a0a9e82d7399ccc52663121 # v4.1.0
+        with:
+          username: ${{ secrets.DOCKER_HUB_USERNAME }}
+          password: ${{ secrets.DOCKER_HUB_TOKEN }}
+
+      - name: Pull nightly images
+        run: docker-compose -f docker-compose-no-volumes.yml pull
```
jayasanka-sack:
```suggestion
        run: docker compose -f docker-compose-no-volumes.yml pull
```

.github/workflows/build-backend.yml
```diff
@@ -145,3 +145,57 @@ jobs:
...
+
+      - name: Start backend and database
+        run: docker-compose -f docker-compose-no-volumes.yml up -d
```
jayasanka-sack:
```suggestion
        run: docker compose -f docker-compose-no-volumes.yml up -d
```

jayasanka-sack:
Alright, now it's working fine. Tested it here: https://github.com/openmrs/openmrs-distro-referenceapplication/actions/runs/24989656579/job/73171764612

.github/workflows/build-backend.yml
jayasanka-sack: This is depending on the build-and-publish job as it pulls from the docker hub. So I'm keeping this.

jayasanka-sack (approved):
Thanks @chibongho!

PR #1010, opened 2026-05-05
O3-5649: Use GitHub Artifacts for spa-assemble-config.json

Bawanthathilan:
# Summary
Update workflow to use GitHub Artifacts for spa-assemble-config.json instead of re-downloading the Docker image.

# Changes
- Extract and upload config as artifact in build job
- Download artifact in deploy job

[openmrs.atlassian.net/browse/O3-5649](https://openmrs.atlassian.net/browse/O3-5649)

### DIFF
--- a/.github/workflows/build-frontend.yml
+++ b/.github/workflows/build-frontend.yml
@@ -67,10 +67,9 @@ jobs:
           IMAGE="${{ env.IMAGE }}:${{ env.DOCKER_IMAGE_TAG }}"
           CONTAINER_ID=$(docker create $IMAGE)
           docker cp $CONTAINER_ID:/usr/share/nginx/html/spa-assemble-config.json ./spa-assemble-config.json
-          docker rm $CONTAINER_ID
 
       - name: Upload spa-assemble-config.json as artifact
-        uses: actions/upload-artifact@v4
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a
         with:
           name: spa-assemble-config
           path: ./spa-assemble-config.json
--- a/.github/workflows/build-frontend.yml
+++ b/.github/workflows/build-frontend.yml
@@ -62,6 +62,20 @@ jobs:
             ${{ env.IMAGE }}:${{ env.DOCKER_IMAGE_TAG_VERSION }}
             ${{ env.IMAGE }}:${{ env.DOCKER_IMAGE_TAG_ENV }}
 
+      - name: Extract spa-assemble-config.json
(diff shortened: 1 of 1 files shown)


.github/workflows/build-frontend.yml
```diff
@@ -62,19 +62,32 @@ jobs:
             ${{ env.IMAGE }}:${{ env.DOCKER_IMAGE_TAG_VERSION }}
             ${{ env.IMAGE }}:${{ env.DOCKER_IMAGE_TAG_ENV }}
 
+      - name: Extract spa-assemble-config.json
+        run: |
+          IMAGE="${{ env.IMAGE }}:${{ env.DOCKER_IMAGE_TAG }}"
+          docker pull $IMAGE
```
Pujan-khunt: I was reviewing this pipeline and had a quick question about this step. Since we just built the Docker image in the same job, could we skip the docker pull and just directly docker create the container from the local cache?

I'm guessing we have to pull it because the `docker/build-push-action` is performing a multi-platform build, which pushes the image directly to the registry and prevents it from loading into the runner's docker daemon? Just wanted to confirm if that was the case.

.github/workflows/build-frontend.yml
```diff
@@ -62,19 +62,32 @@ jobs:
...
+          CONTAINER_ID=$(docker create $IMAGE)
+          docker cp $CONTAINER_ID:/usr/share/nginx/html/spa-assemble-config.json ./spa-assemble-config.json
+          docker rm $CONTAINER_ID
+
+      - name: Upload spa-assemble-config.json as artifact
+        uses: actions/upload-artifact@v4
+        with:
+          name: spa-assemble-config
+          path: ./spa-assemble-config.json
```
Pujan-khunt, line 75: Since the `spa-assemble-config.json` file is only needed for the next job, it only needs to be stored temporarily. We can limit it using `retention-days: 1`. Is this is a valid approach?

.github/workflows/build-frontend.yml
jayasanka-sack: Yeah, no need to pull. We should use the image built by the job.

.github/workflows/build-frontend.yml
jayasanka-sack: yes, 1 day would be more than enough.

Bawanthathilan: Thanks i will update my PR @Pujan-khunt

jayasanka-sack (approved):
Thanks @Bawanthathilan!

PR #1016, opened 2026-05-14
Add task-list backend and frontend modules to the distro

denniskigen:
Adds the task-list feature to the reference application distro by including both the backend OMOD (`tasks-omod`) and the frontend ESM (`@openmrs/esm-task-list-app`).

### Backend (`distro/`)

- `pom.xml`: declare `tasks.version` property (`1.0.0-SNAPSHOT`) and add a `tasks-omod` dependency under `org.openmrs.module`.
- `distro.properties` and `distro-no-demo.properties`: add `omod.tasks=${tasks.version}` so the SDK build picks up the OMOD.

### Frontend

- `frontend/spa-assemble-config.json`: add `@openmrs/esm-task-list-app` to the `frontendModules` map, pinned to `next` to match the convention used by every other module here.

### Notes

- `tasks-omod` is currently `1.0.0-SNAPSHOT` — no release has been published yet. CI will need the SNAPSHOT artifact available on `mavenrepo.openmrs.org/snapshots`, or this should be repinned to a release once one is cut.
- The frontend ESM resolves via the import-map service against the `next` dist-tag, same as every other module in this file.
- Context: this came up while looking at why the task-list e2e tests in [openmrs-esm-patient-chart#3303](https://github.com/openmrs/openmrs-esm-patient-chart/pull/3303) were failing — the feature isn't loaded in that repo's e2e env either. Adding task-list here doesn't fix that PR (patient-chart builds its own e2e image), but it does make task-list a first-class part of the refapp.

### DIFF
--- a/distro/distro-no-demo.properties
+++ b/distro/distro-no-demo.properties
@@ -33,4 +33,5 @@ omod.event=${event.version}
 omod.bedmanagement=${bedmanagement.version}
 omod.stockmanagement=${stockmanagement.version}
 omod.billing=${billing.version}
+omod.tasks=${tasks.version}
 content.referenceapplication=${reference-content.version}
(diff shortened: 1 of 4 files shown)


wikumChamith (approved):

NethmiRodrigo (approved):
Thanks @denniskigen!

PR #1020, opened 2026-05-20
(fix) Use task list backend snapshot

denniskigen:
## Summary

Updates the reference application distro to include `tasks-omod` `1.1.0-SNAPSHOT`. Reviewing `openmrs-module-tasks` showed that this snapshot supports the patient task list app rationale payload in `activity.detail.reasonCode`, so it is the backend version needed by the current task-list app contract.

This pairs with openmrs/openmrs-esm-patient-chart#3316, which lowers the task-list app dependency from the unreleased `tasks >=2.0.0` requirement to `tasks >=1.1.0-SNAPSHOT`. Once the backend image rebuilds with this distro change, openmrs/openmrs-esm-core#1779 can add `@openmrs/esm-patient-task-list-app` to the core E2E environment without implementer tools reporting unresolved backend dependencies.

### DIFF
--- a/distro/pom.xml
+++ b/distro/pom.xml
@@ -55,7 +55,7 @@
     <!-- Stock Management and Billing -->
     <stockmanagement.version>3.0.0</stockmanagement.version>
     <billing.version>2.3.0-SNAPSHOT</billing.version>
-    <tasks.version>1.0.0</tasks.version>
+    <tasks.version>1.1.0-SNAPSHOT</tasks.version>
     <!-- Content Packages -->
     <reference-content.version>1.4.0</reference-content.version>
     <reference-demo-content.version>1.8.0-SNAPSHOT</reference-demo-content.version>


denniskigen:
Closing this as well following @ibacher's clarification that the Tasks module change is intended to ship as `2.0.0` because it is a breaking API change.

NethmiRodrigo:
@denniskigen @ibacher since the tasks module’s [current version appears to be 2.1.0](https://github.com/openmrs/openmrs-module-tasks/blob/22e0fa5f313c0aa42c7249564e5435f2e6266600/pom.xml#L7), shouldn’t we just update the distro to use that version instead?

denniskigen:
Has the release been cut yet? I don't see an updated tag, @NethmiRodrigo.

NethmiRodrigo:
> Has the release been cut yet? I don't see an updated tag, @NethmiRodrigo.

@denniskigen hmm yeah, I don’t see one either but I think whatever version in the pom.xml has to be a version that is released and available.

PR #1028, opened 2026-05-25
fix(O3-4977): Unwanted referenceapplication-demo folder in RefApp 3.4.0 build

Josephkagimu1:
Added `content.referenceapplication-demo.namespace=referenceapplication` to `distro.properties`. This tells the SDK to use `referenceapplication` as the namespace, merging the demo content into the correct domain folder.

Before :
Screenshot 2025-08-06 at 6 42 07 PM (1)

After :
hh

Verification :
1. Run `mvn clean package` in the `distro/` directory
2. Run `jar tf target/distro-emr-configuration-3.7.0-SNAPSHOT.zip | findstr "referenceapplication-demo"`
3. No output confirms , the unwanted folder is therefore removed .

Ticket : https://openmrs.atlassian.net/browse/O3-4977

### DIFF
--- a/distro/distro.properties
+++ b/distro/distro.properties
@@ -37,3 +37,4 @@ omod.billing=${billing.version}
 omod.tasks=${tasks.version}
 content.referenceapplication=${reference-content.version}
 content.referenceapplication-demo=${reference-demo-content.version}
+content.referenceapplication-demo.namespace=referenceapplication


Josephkagimu1:
Please check on this PR to confirm , I managed to fix this @denniskigen  @jayasanka-sack , @ibacher

solomonfortune (commented):
hey @Josephkagimu1 could you please review the title format of your PR. the right format is (fix) O3-4977

ibacher:
This isn't unwanted. The data intentionally comes from two different content packages and we want each content package to be namespaced so we don't accidentally overwrite existing files. These two content packages do not and should not need to coordinate file names.

PR #1029, opened 2026-05-27
Add patient procedures app to configuration

jayasanka-sack:
The goal of this PR is to add the newly created procedure history frontend app to the distro.


jayasanka-sack:
cc: @dilankavishka

PR #1036, opened 2026-06-04
Update reference-demo-content.version to 1.8.0

wikumChamith:


jayasanka-sack:
Curious, can't we use the release job to achieve this?

wikumChamith:
This content package still doesn't have a GHA for release.

dkayiwa:
> This content package still doesn't have a GHA for release.

Is now the time to add it?

wikumChamith:
> > This content package still doesn't have a GHA for release.
> 
> Is now the time to add it?

We need to check if we can reuse our parent module release GHA for content packages as well.

NethmiRodrigo (approved):

ibacher:
> We need to check if we can reuse our parent module release GHA for content packages as well.

It probably needs a Java version specified, but the GHA stuff is pretty generic as long as something uses the maven-releases-plugin for releases.

jayasanka-sack:
IIRC we need to update the matrix for 1.x as content packages support
older versions too.

wikumChamith:
> > We need to check if we can reuse our parent module release GHA for content packages as well.
> 
> It probably needs a Java version specified, but the GHA stuff is pretty generic as long as something uses the maven-releases-plugin for releases.

We can use the release GHA, but I don't think we'll be able to use the one for the build, as it looks for an omod file in some steps.

PR: https://github.com/openmrs/openmrs-content-referenceapplication-demo/pull/75

PR #1051, opened 2026-07-10
Rework the workflows for deployments, QA and releases

ibacher:
This is a large series of changes to the workflows for this repo primarily aimed at replicating behaviour from Bamboo that was not fully ported over while taking advantage of the flexibility we get from GitHub Actions versus Bamboo. 

Key points:

### Deployment Jobs

There are four environments we deploy to:

* dev3.openmrs.org
* sec3.openmrs.org (a mirror of dev3)
* test3.openmrs.org
* o3.openmrs.org

All of these deploy jobs have 2 or 3 responsibilities:

* Update the server when new artifacts are available (on every update for dev3 / sec3 and every relevant release for the other two)
* Allow us to run them (via workflow_dispatch) to redeploy or reset the server
* For test3 and o3 only, a scheduled reset of the server at the appropriate interval

These jobs have been "standardized" so that they are basically the same overall flow with only minimal necessary differences.

### Build Frontend

In the Bamboo setup, each prerelease (on the appropriate repo) would trigger a webhook call to rebuild the frontend once the artifacts were published. The current job here only triggered on a cron schedule which is a bad fit for continuous integration. This PR adds a `repository_dispatch` trigger that will be used to more closely mirror the Bamboo setup.

### Release Jobs

The release jobs are mostly the same as before with some reordering and a few fixes (e.g., leveraging the right job to run E2E tests for "releases"). I also changed things to use the Maven versions plugin to update versions, as this behaves more correctly than sed does, especially with ensuring that versions are properly in sync between the different artifacts. Finally, I restructured things a little bit to ensure we are always building all 5 images for all releases, since we have been releasing (and using!) the monitoring and certbot images.

I removed from the release job the check against SNAPSHOTS. This is not a concern that should be handled solely in CI. Rather we should make those restrictions that we want to enforce part of the Maven build, like we do everywhere else using the Maven Enforcer plugin. Ideally, our CI jobs should be fairly thin orchestration layers around tasks defined elsewhere, because this ensures we can run those sort of checks in multiple contexts.

### Runbook

There was a release runbook, but it was buried in the `.github/` directory and contained a bunch of stuff that really probably belongs on the Wiki. I've moved a reduced and slightly more focused runbook to be a first-class piece of documentation in this repo.


RELEASE.md
```diff
@@ -0,0 +1,63 @@
...
+gated to the `openmrs` org and requires the `DEPLOY_SSH_KEY` / `SERVER_HOST` secrets to be set.
+
+## Environments
+
+| Env   | Host              | Images tag | How it deploys                                              |
+|-------|-------------------|------------|------------------------------------------------------------|
+| dev3  | dev3.openmrs.org  | `dev`      | Auto on every successful `main` image build; also manual. Also deploys sec3. |
+| test3 | test3.openmrs.org | `qa`       | By *Cut QA Release Candidate*; also manual; weekly reset (Mon 01:00 UTC). |
+| o3    | o3.openmrs.org    | `demo`     | By *Promote QA to Prod*; also manual; nightly reset (01:00 UTC). |
+
```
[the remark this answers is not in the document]
ibacher: The runbook here is correct. The fact that this server is "bonga" is not a fact that should be recorded long-term as we can and do move servers.

.github/workflows/e2e-on-release.yml
```diff
@@ -23,21 +37,23 @@ jobs:
...
         run: while [[ "$(curl -s -o /dev/null -w ''%{http_code}'' http://localhost/openmrs/login.htm)" != "200" ]]; do sleep 10; done
 
+      # The tree at a release ref pins modules to a floating dist-tag
+      # ("latest"/"next"), so the concrete versions to test against only exist
+      # in the resolved manifest baked into the built frontend image.
+      - name: Resolve pinned module versions from the built frontend image
+        id: refs
+        run: |
+          docker cp frontend:/usr/share/nginx/html/spa-assemble-config.json frontend/spa-assemble-config.json
+          bash ./tests/e2e/extract_tag_numbers.sh
```
dkayiwa, line 55: The manifest this copies out of the image is minified single-line JSON (`openmrs assemble --manifest` writes it that way), and `extract_tag_numbers.sh` parses it with `awk -F'"' '$0 ~ app {print $4}'`, which on a one-line file prints field 4 of the whole line, i.e. the coreVersion, for every module. I ran the script against the real manifest from the published `distro-emr-frontend:3.7.1-rc.4` zip (the same file this `docker cp` produces) and all six outputs came back `refs/tags/v10.0.0`. Merged as-is, every release E2E run goes wrong in two ways: form-builder, cohort-builder and dispensing fail at checkout (no v10.0.0 tag exists there), while patient-chart and patient-management do have v10.0.0 tags, so those two would silently run suites two major versions older than the RC under test. This needs fixing before merge; pretty-printing the manifest before handing it to the script restores the one-entry-per-line shape it expects:

```suggestion
          docker cp frontend:/usr/share/nginx/html/spa-assemble-config.json /tmp/qa-manifest.json
          python3 -m json.tool /tmp/qa-manifest.json > frontend/spa-assemble-config.json
          bash ./tests/e2e/extract_tag_numbers.sh
```

Longer term I'd make extract_tag_numbers.sh parse the JSON with python or jq so formatting can't break it, and give it a small fixture test that feeds it a minified manifest; nothing in the current suite or actionlint can catch this class of breakage.

.github/workflows/release-qa.yml
```diff
@@ -125,43 +117,40 @@ jobs:
...
-            if [ "$n" != "1" ]; then
-              echo "::error::$f: expected exactly one occurrence of <version>$OLD</version>, found $n"; exit 1
-            fi
-            sed -i "s|<version>${OLD_ESC}</version>|<version>${NEW}</version>|" "$f"
+            grep -q "<version>${NEW}</version>" "$f" \
+              || { echo "::error::$f is not at $NEW after versions:set"; exit 1; }
           done
           git diff --stat
 
-      - name: Refuse SNAPSHOT or timestamp-locked module versions
```
dkayiwa, line 157: I'd keep this guard until its Maven-side replacement actually exists. The enforcer rule isn't in this PR, and main already carries the case this guard exists for: `reference-demo-content.version` was bumped to `1.9.3-SNAPSHOT` after this branch was cut, and it flows into `distro.properties` as `content.referenceapplication-demo`. So merged as-is, the next RC cut from main pins the demo content package to a floating snapshot with nothing refusing it, in CI or in the build, and every image rebuild can resolve different bits than QA signed off on. That is the same silent failure that shipped 3.7.0 broken. I agree the check belongs in the Maven build; one wrinkle for that follow-up is that `requireReleaseDeps` (with `onlyWhenRelease`) covers the omod dependencies but not the content-package version properties that only surface in `distro.properties`, which this grep did cover. Landing the enforcer rules in this PR, or dropping this step in the PR that adds them, would keep that window closed.

.github/workflows/deploy-o3.yml
```diff
@@ -1,171 +1,54 @@
...
-permissions: {}
+  schedule:
+    # 1 AM UTC = 4 AM EAT = 6:30 AM IST
+    - cron: '0 1 * * *'
 
 concurrency:
-  # ':' cannot appear in a git ref, keeping this outside any ref-derived
-  # namespace other workflows might use.
-  group: deploy-o3::production
+  group: deploy-o3
```
dkayiwa: This group only serializes standalone runs of this workflow. When release-prod calls it via workflow_call, the jobs run inside the caller's run, which already holds the `o3-release` group, and a workflow-level `concurrency` in a called workflow doesn't attach to that run (the documented mechanism for reusable workflows is job-level concurrency, and `github.workflow` in here resolves to the caller's name anyway). The practical gap: a promote whose deploy lands around 01:00 UTC can SSH concurrently with the scheduled reset, so `deploy emr-3-demo --destroy-volumes` and `deploy emr-3-demo` race on the same target. The pre-PR files serialized the SSH step with a job-level group for exactly this reason. Moving the group onto the `deploy` job (here and in deploy-test3.yml, which has the same shape with its Monday reset) makes the serialization hold for both the standalone and the called paths:

```yaml
jobs:
  deploy:
    concurrency:
      group: deploy-o3
      cancel-in-progress: false
```

If the server-side `deploy` script already locks per target this is belt and braces, but it costs nothing and keeps the guarantee visible in the workflow.

.github/workflows/release-qa.yml
```diff
@@ -1,20 +1,4 @@
...
-#   - frontend/Dockerfile: APP_SHELL_VERSION next -> <core_version>
-#   - the three project poms -> <version>-rc.1
-# Later runs for the same version add rc.2, rc.3, ... on the existing branch.
-#
-# It then dispatches the existing image build workflows on the release branch,
-# publishing images tagged <version>-rc.N and qa (which test3.openmrs.org runs).
-#
-# This workflow never pushes to main. When a release makes the development
-# version on main stale (minor releases), it opens a version-bump PR instead.
+name: 'Release: Cut QA Release Candidate'
```
dkayiwa, line 1: Nothing in this workflow calls the Actions API anymore now that the image builds and E2E moved from `gh workflow run` dispatches to `workflow_call` jobs, so the `actions: write` in the permissions block below can be dropped (release-prod already dropped it in this PR).

.github/workflows/release-qa.yml
ibacher: Yeah, I'm leaving that as a separate PR. I actually think that what this check does is substantially more _harmful_ than what it's trying to prevent. More specifically, the check here is too aggressive and completely ignores the reasons that we had versions pinned the way we do.

.github/workflows/deploy-o3.yml
ibacher: That seems correct, but this isn't really a new addition in this PR, so let's make that a follow-up point (again, the scope of this PR is big enough, I think).

.github/workflows/e2e-on-release.yml
ibacher: This has been fixed, albeit by moving to `jq` in the script itself, trying to keep these thing wrappers.

.github/workflows/release-qa.yml
ibacher: Fixed

PR #1052, opened 2026-07-15
Added auditlogweb module to distro

druchniewicz:


druchniewicz:
@wikumChamith I think we can add [auditlogweb](https://github.com/openmrs/openmrs-module-auditlogweb) module to distro. Here's PR

wikumChamith:
@jayasanka-sack, @dkayiwa, @ibacher, what do you think?

dkayiwa:
@wikumChamith is it mature enough?

wikumChamith:
> @wikumChamith is it mature enough?

@dkayiwa, I think this module is now at a stage where we need feedback from the wider community for it to mature further. After GSoC ends, I believe this module will be able to handle auditing the most important events, including patient record reads.

dkayiwa:
@wikumChamith it feels like it is that feedback from the wider community that will determine whether we bundle it in the distro or not.

wikumChamith:
> @wikumChamith it feels like it is that feedback from the wider community that will determine whether we bundle it in the distro or not.

The problem is that this creates a chicken-and-egg situation: we wait to get feedback from the community, but we don't get any feedback because it's not in the distro.

jayasanka-sack:
Quick note on why I think auditlogweb can't go in the refapp yet.

The blocker is obs. Obs is `@Audited` in core. So with Envers on, every obs write also writes an obs_aud row. That's basically 1:1 with the obs table, which is already our biggest growth concern. So turning this on doubles it. And there's no clean per-entity off switch, Envers is all or nothing behind the global flag. An implementer who wants audit for everything except obs has no supported way to get there short of patching core.

Annoying part is we don't really need it for obs anyway. Obs are immutable, edits void the old one and create a new one with a previousVersion pointer, so the full history is already on the table. Enversing it just duplicates that.

Rough options: simplest is dropping `@Audited` from the high volume clinical stuff (obs, encounter, order) and keeping Envers for low volume metadata (users, roles, privileges, global props, concepts). Perhaps we can use the existing audit columns for obs, etc. Bigger fix is a runtime allowlist so implementers choose what gets audited. Heavy clinical auditing should go to an external pipeline. not core.

tl;dr; keep it standalone for now, do the annotation trim for refapp, raise the allowlist as the longer term fix. Lmk what you think.

ibacher:
So, I have several issues with the audit module:

1. I don't think the default opt-everything-in implementation is the right way to do this, especially when we have a number of tables that are known to already grow to problematic sizes (obs, OCL imports, Conditions, Orders, basically anything we treat as "mostly append"). This seems like adding this in and turning on envers will likely only increase performance issues.
2. I don't think the Envers solution has been properly tuned to work with our "immutable" entities, which are "immutable" precisely for the sort of historical auditing. This still needs to be rationalized and worked through.
3. As I understand things, we still don't have auditing of the most important thing which is records that are viewed.

I'm willing to talk about spinning up a different server to showcase the audit work, but it is not ready to be included in the RefApp.

druchniewicz:
Thanks for all your comments and quick responses

jayasanka-sack:
Thanks for opening the PR and raising this @druchniewicz! ❤️ As Ian suggested having a separate server to gather feedback seems like a good immediate next step.

PR #1054, opened 2026-07-20
Add container and JVM metrics and alerting to the monitoring stack

jayasanka-sack:
This PR adds container and JVM metrics, and alerts to the monitoring stack.

- **cAdvisor** exposes per-container metrics (CPU, memory, etc.) and Prometheus scrapes it.
- **Prometheus JMX exporter** runs as a javaagent on the backend (attached via `OMRS_JAVA_SERVER_OPTS`, scraped on `9404`) to expose JVM metrics. The jar is downloaded into a shared volume during monitoring init and reused on later starts.
- **Grafana-managed alerting** handles alerts, provisioned from `grafana/alerting/rules.yaml`. 
- **Grafana** shows it all through the container, JVM, endpoint health check, and logs dashboards.

### Backend container dashboard
image

### JVM Runtime Dashboard
image

### Alert Rules
image


monitoring/entrypoint.sh
```diff
@@ -20,14 +20,35 @@ cp /monitoring/grafana-datasources.yaml /etc/grafana/provisioning/datasources/da
...
 mkdir -p /config
 cp /monitoring/blackbox.yml /config/blackbox.yml
 
-echo "Configuration initialization complete."
\ No newline at end of file
+echo "Setting up jmx exporter..."
+JMX_EXPORTER_VERSION=1.6.0
+JMX_AGENT_JAR="jmx_prometheus_javaagent-${JMX_EXPORTER_VERSION}.jar"
+JMX_AGENT_URL="https://github.com/prometheus/jmx_exporter/releases/download/v${JMX_EXPORTER_VERSION}/${JMX_AGENT_JAR}"
+mkdir -p /jmx-exporter-data/
```
ibacher, line 41: It seems like these steps could be done in the Dockerfile, no?

docker-compose.grafana.yml
```diff
@@ -77,6 +78,34 @@ services:
...
+      container_name: cadvisor
+      privileged: true
+      devices:
+        - /dev/kmsg
+      volumes:
+        - /:/rootfs:ro
+        - /sys:/sys:ro
+        - /var/lib/docker/:/var/lib/docker:ro
+        - /run/containerd/containerd.sock:/run/containerd/containerd.sock:ro
+        - ${HOME}/.docker/run/docker.sock:/var/run/docker.sock:ro
```
dkayiwa: This socket path only exists on Docker Desktop for Mac. On the Linux hosts this actually runs on (dev3, plus anyone who copies the overlay per the README), `${HOME}/.docker/run/docker.sock` doesn't exist, Docker quietly creates it as an empty directory, and cadvisor can't reach the Docker API. With `--docker_only=true` that means no per-container series at all, so the new dashboards stay blank and, because the container alerts use `noDataState: OK`, ContainerOOMKilled, DbHighMemory and friends sit green forever. TargetDown won't catch it either, since cadvisor's own /metrics endpoint keeps serving fine. The alloy service above already mounts the socket the standard way, and that path also works on Docker Desktop:

```suggestion
        - /var/run/docker.sock:/var/run/docker.sock:ro
```

docker-compose.grafana.yml
```diff
@@ -77,6 +78,34 @@ services:
...
+      command:
+        - --docker_only=true
+        - --store_container_labels=true
+        - --docker=unix:///var/run/docker.sock
+
+    backend:
+      environment:
+        OMRS_JAVA_SERVER_OPTS: "-javaagent:/opt/jmx_exporter/jmx_prometheus_javaagent-1.6.0.jar=9404:/opt/jmx_exporter/jmx_config.yml"
```
dkayiwa: Setting this replaces the image's default server opts instead of adding to them: startup-init.sh in openmrs-core only applies `-Dfile.encoding=UTF-8 -server -Djava.security.egd=file:/dev/./urandom -Djava.awt.headless=true -Djava.awt.headlesslib=true` when the variable is unset (https://github.com/openmrs/openmrs-core/blob/2.8.x/startup-init.sh#L37), and startup.sh assigns it straight into JAVA_OPTS. So enabling the monitoring overlay silently changes the backend's JVM flags. Nothing crashes today on corretto 21 (UTF-8 is the default charset since JDK 18, and headless auto-detects when DISPLAY is unset), but monitoring shouldn't alter app flags as a side effect. I'd repeat the defaults and append the agent:

```suggestion
        OMRS_JAVA_SERVER_OPTS: "-Dfile.encoding=UTF-8 -server -Djava.security.egd=file:/dev/./urandom -Djava.awt.headless=true -Djava.awt.headlesslib=true -javaagent:/opt/jmx_exporter/jmx_prometheus_javaagent-1.6.0.jar=9404:/opt/jmx_exporter/jmx_config.yml"
```

docker-compose.grafana.yml
```diff
@@ -77,6 +78,34 @@ services:
...
+      volumes:
+        - jmx-exporter-data:/opt/jmx_exporter:ro
+      ports:
+        - "9404:9404"
```
dkayiwa: Prometheus scrapes `backend:9404` over the compose network, so publishing the port on the host isn't needed for anything in this PR. It would also be the only host-published port in the whole monitoring stack (Grafana itself is only reachable through the gateway's `/grafana` route). Published like this it binds 0.0.0.0, and Docker's iptables rules bypass ufw-style host firewalls, so if this ships to an internet-facing host like dev3, the full JMX dump (every JVM and Tomcat MBean, given the catch-all `pattern: ".*"` config) is readable by anyone, no auth. Please drop the mapping; if you want it for host-side debugging, `127.0.0.1:9404:9404` keeps it off the network.

```suggestion
```

docker-compose.grafana.yml
```diff
@@ -77,6 +78,34 @@ services:
             monitoring-init:
                 condition: service_completed_successfully
 
+    cadvisor:
+      image: gcr.io/cadvisor/cadvisor:latest
```
dkayiwa: Every other image in this file is pinned (loki 3.6, alloy v1.13.1, prometheus v3.9.0, blackbox v0.28.0, grafana 12.3); cadvisor is the only `latest`. It's also not buying freshness: gcr.io's `latest` currently resolves to v0.55.1, and the actual latest upstream release (v0.60.5) isn't published to gcr.io at all. Pinning what you tested keeps a silent tag move from changing metric names under the dashboards:

```suggestion
      image: gcr.io/cadvisor/cadvisor:v0.55.1
```

monitoring/entrypoint.sh
dkayiwa: Agreed, and I think it matters beyond tidiness: this PR also gives backend a `depends_on` on monitoring-init, and the script runs under `set -e`, so on a fresh environment (empty volume) a failed download from GitHub keeps the whole EMR from starting, not just the monitoring stack. GitHub release downloads do intermittently fail with 403s from rate limiting. If the jar is baked into the image with a `RUN wget` in the Dockerfile (CI already builds and pushes this image), a bad download fails the CI build loudly instead, and the entrypoint just copies the jar into the volume the same way it already copies jmx_config.yml. The download-and-cache logic here then goes away entirely.

monitoring/grafana/alerting/rules.yaml
```diff
@@ -0,0 +1,675 @@
...
+  - orgId: 1
+    name: ssl
+    folder: OpenMRS Alerts
+    # TLS certificate expiry moves on the scale of days, not minutes, so this
+    # rule gets its own group with a daily evaluation interval rather than
+    # sharing the 1m interval of the availability group.
+    interval: 24h
+    rules:
+      - uid: openmrs-ssl-cert-expiring-soon
+        title: SSLCertExpiringSoon
```
ibacher, line 133: Prometheus only reads data from the Gateway using `http`, so can this rule ever fire?

monitoring/grafana/alerting/rules.yaml
```diff
@@ -0,0 +1,675 @@
...
+        # evaluation, so any non-zero value here would delay firing by a whole
+        # day. A single reading of cert expiry is authoritative — unlike a
+        # health probe, it cannot flap.
+        for: 0m
+        labels:
+          severity: warning
+        annotations:
+          summary: "TLS certificate for {{ $labels.instance }} expires soon"
+          description: "The TLS certificate for {{ $labels.instance }} expires in {{ $values.B.Value | humanizeDuration }}."
+        noDataState: OK
```
ibacher, line 145: It feels like `noDataState` should be different than `OK`, depending on different variables. E.g., if we're not using SSL, it's fine to have no data on this rule. But if we are, it's wrong. This is also set for all other rules and that feels like a wrong default.

docker-compose.grafana.yml
ibacher: Yeah, actually, this probably makes sense to introduce a new environment variable like `OMRS_JAVA_SERVER_OPTS_EXT`, because actually disabling those defaults put things in a much worse position, especially missing the `headless` and `egd` settings.

ibacher: One thing I want to point out here is that this requires the Java Agent to _always_ be version 1.6.0. It seems like in the build process, we should be downloading the specified version but copying it to a file like `/openmrs/openmrs_agents/jmx_prometheus_javaagent.jar`, so that upgrading the version is less fragile.

monitoring/Dockerfile
```diff
@@ -1,9 +1,18 @@
 FROM alpine:3.23
 
+# ca-certificates lets busybox wget download the JMX exporter jar over HTTPS
+RUN apk add --no-cache ca-certificates
+
+# Bake the JMX exporter agent into the image
+ARG JMX_EXPORTER_VERSION=1.6.0
+RUN mkdir -p /opt/jmx-exporter \
+    && wget -q -O "/opt/jmx-exporter/jmx_prometheus_javaagent-${JMX_EXPORTER_VERSION}.jar" \
```
ibacher, line 9: I think we've preferentially been using CURL. Not a _big_ deal, but it's nice to be consistent.

docker-compose.grafana.yml
ibacher: Try to be the last comment I make here, but also overriding this in the docker-compose.yml environment seems doubly-dangerous. The fact that this is necessary isn't documented; if `OMRS_JAVA_SERVER_OPTS` is overridden (e.g., by an `.env` file, this doesn't get deployed, etc.

The more I think about it, the more I think we want, in the core Docker images, to support `/openmrs/openmrs_agents` and loading those as part of the startup scripts rather than purely doing this here.

docker-compose.grafana.yml
```diff
@@ -77,6 +78,32 @@ services:
...
+        - --docker=unix:///var/run/docker.sock
+
+    backend:
+      environment:
+        OMRS_JAVA_SERVER_OPTS: "-javaagent:/opt/jmx_exporter/jmx_prometheus_javaagent-1.6.0.jar=9404:/opt/jmx_exporter/jmx_config.yml"
+      volumes:
+        - jmx-exporter-data:/opt/jmx_exporter:ro
+      depends_on:
+        monitoring-init:
+          condition: service_completed_successfully
```
ibacher, line 116: It feels like making `monitoring-init` a hard dependency here is the wrong thing to do. Maybe mark this as non-required. Or just use `volumes_from:` since that's really what we're kind of depending on.

monitoring/grafana/dashboards/container-dashboard.json
```diff
@@ -0,0 +1,260 @@
...
+      "type": "timeseries",
+      "title": "CPU Usage",
+      "description": "CPU consumed by the backend container, as a percentage of a single core (100% = one full core; can exceed 100% on multi-core hosts). Watch for sustained highs or a rising baseline. If it sits near the host's core limit during normal use, allocate more CPU or scale out; brief spikes during startup or batch jobs are usually fine. Correlate spikes with slow requests or high load.",
+      "gridPos": { "h": 8, "w": 12, "x": 0, "y": 0 },
+      "datasource": { "type": "prometheus", "uid": "Prometheus" },
+      "targets": [
+        {
+          "refId": "A",
+          "datasource": { "type": "prometheus", "uid": "Prometheus" },
+          "expr": "sum by (name) (rate(container_cpu_usage_seconds_total{name=\"backend\",cpu=\"total\"}[5m])) * 100",
```
ibacher, line 21: So `rate(container_cpu_usage_seconds_total)` gives us the amount of CPU cores used. With only one CPU, 0.9 = 0.9 * 1 (cores) = 90% usage. With _multiple_ cores, however, this breaks, since 8 CPUs running at 0.11 = 0.11 * 8 = 88% usage, even though the "real" usage is 11%.

```
sum by (name) (rate(container_cpu_usage_seconds_total{name!=""}[5m])) / on() group_left() machine_cpu_cores
```

Should fix the issue by giving you total core usage as a percentage of total cores.

ibacher: Note that this `rate(container_cpu_usage_seconds_total)` is used in both rules and dashboards and this issue applies to all usages.

ibacher: I'm less sure about this one, but Claude flagged for me that `cpu="total"` will stop working if per-CPU metrics are turned on and that if we just remove that filter things should continue to work? Not convinced, just flagging.

monitoring/jmx_config.yml
ibacher, line 1: Claude pointed out to me that several MBeans will export information we _don't_ want exported to a monitoring tool, e.g., if passwords are stored as readable MBeans. It might be better to filter this down, e.g., to `[jvm_*, process_*]`.

docker-compose.grafana.yml
```diff
@@ -77,6 +78,32 @@ services:
             monitoring-init:
                 condition: service_completed_successfully
 
+    cadvisor:
+      image: gcr.io/cadvisor/cadvisor:v0.55.1
+      container_name: cadvisor
+      privileged: true
```
ibacher, line 84: So the fact that we're running a privileged container should be explained in the README and flagged. (Really, this whole monitoring feature needs to be better documented).

docker-compose.grafana.yml
```diff
@@ -77,6 +78,32 @@ services:
...
+        - /dev/kmsg
+      volumes:
+        - /:/rootfs:ro
+        - /sys:/sys:ro
+        - /var/lib/docker/:/var/lib/docker:ro
+        - /run/containerd/containerd.sock:/run/containerd/containerd.sock:ro
+        - /var/run/docker.sock:/var/run/docker.sock:ro
+      command:
+        - --docker_only=true
+        - --store_container_labels=true
```
ibacher, line 95: This actually probably grabs too much, since this container has access to _all_ containers on the host machine not just this docker-compose stack. We should be using whitelisting and probably some kind of stack label to ensure we're only grabbing info on these containers.

docker-compose.grafana.yml
```diff
@@ -77,6 +78,32 @@ services:
             monitoring-init:
                 condition: service_completed_successfully
 
+    cadvisor:
```
ibacher, line 81: nit: we've used 4-space indentation in this file. Please follow the existing convention.

monitoring/grafana/alerting/rules.yaml
```diff
@@ -0,0 +1,675 @@
...
+
+      - uid: openmrs-db-high-memory
+        title: DbHighMemory
+        condition: C
+        for: 5m
+        labels:
+          severity: warning
+        annotations:
+          summary: "MariaDB memory usage is high"
+          description: "The db container's working-set memory has been above 2 GiB for more than 5 minutes. TUNABLE: adjust the threshold to your host, or add a mem_limit to the db service so this can become a percentage-of-limit alert."
```
ibacher, line 577: How can someone tune this?

monitoring/Dockerfile
ibacher: One more thing: this whole JMX setup seems really complicated. We download and install it in one place on this file, so it can be copied to a volume location which can be conditionally mounted in the backend image, where it can be read and used. This feels like a overly complicated song and dance to get a static file into the container. It might be better to make the JMX exporter just a standard part of the backend image, but only enable it via an environment variable or the like.

ibacher (commented):
So a bunch of comments, one thing I probably mislead you because I wasn't really thinking through everything: this heavily builds on Prometheus for gathering telemetry, etc., but we're already leveraging Alloy for logs, but Alloy can also consume OpenTelemetry data which is the tech stack we want to leverage for telemetry as part of this project. It's probably slightly more inline with that part of the vision if we published things from the backend -> Alloy -> Grafana instead of routing logging through Alloy and everything else through Prometheus.

ibacher:
Replace by #1069

PR #1055, opened 2026-07-24
Add custom reports module and Stock Management fixes

anmolpawar121:
Adds a new labtestreport OpenMRS module (classic admin UI + O3 frontend + Reports dashboard registration) with five interactive reports: Lab Test Summary, Patient Encounter Summary, Disease Surveillance, Session Attendance, and a Stock Inventory Ledger - each with drill-down to the underlying patients where applicable, plus KPI tiles, search/category filters, and a table/graph toggle on the O3 side.

Also forks and patches the Stock Management backend module and its O3 frontend to fix several real bugs blocking normal use: stock operations never sent a location and the backend whitelist rejected it when they did, the Role Scope "include descendants" toggle had no onChange handler, the dashboard's "Out of stock"/"Expiring stock" widgets computed the wrong numbers entirely, and the non-drug item picker searched the whole concept dictionary instead of just non-drug items (now backed by a new /non-drug REST resource).

Wires both new/patched modules into the root Dockerfile, pom.xml, distro properties, and frontend assemble config so they ship with the app.


anmolpawar121:
s

PR #1053, opened 2026-07-16
(chore) Bump patientdocuments to 1.2.0-SNAPSHOT for visit summary support

BlessedAmrita:
## Summary

Bumps the patientdocuments module from 1.1.0 to 1.2.0-SNAPSHOT so the reference application ships the version that contains the visit summary PDF pipeline and its REST endpoint.

## Related

Part of https://openmrs.atlassian.net/browse/O3-5667

PR #1058, opened 2026-08-12
Feat: bump patientdocuments visit summary

BlessedAmrita:
## Summary

Adds `frame-src 'self' blob:;` to the gateway CSP so blob-URL iframes under `/openmrs/spa/` can load. The patientdocuments visit-summary preview renders a generated PDF into a blob iframe; with no `frame-src`, the browser fell back to `default-src` (which lacks `blob:`) and blocked it.

## Changes

`frame-src 'self' blob:;` added to the `default` branch of the `$csp_header` map in both `default.conf.template` and `default-ssl.conf.template`. Other branches and `frame-ancestors` unchanged.

## Testing

`nginx -t` passes on the rendered template; `envsubst` leaves `blob:` and `${FRAME_ANCESTORS}` intact. Live confirmation once dev3 rebuilds the gateway.

### DIFF
--- a/distro/pom.xml
+++ b/distro/pom.xml
@@ -28,7 +28,7 @@
     <fhir2.version>4.2.0</fhir2.version>
     <webservices.rest.version>3.5.0</webservices.rest.version>
     <addresshierarchy.version>2.21.0</addresshierarchy.version>
-    <patientdocuments.version>1.1.0</patientdocuments.version>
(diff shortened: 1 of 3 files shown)


BlessedAmrita:
Superseded by #1059, the version bump here is already on main via #1053, so this only left the CSP change, which I've moved to its own PR.

PR #1059, opened 2026-08-12
(fix) allow blob: object-src so SPA <object> PDF previews can load

BlessedAmrita:
### Summary
  Adds `object-src 'self' blob:;` to the gateway CSP so blob-URL `<object>` embeds a generated PDF into a blob `<object>`; with no `object-src`, the browser fell back to `default-src` (which lacks `blob:`) and blocked it.

  ### Changes
  `object-src 'self' blob:;` added to the default branch of the `$csp_header` map in both `default.conf.template` and `default-ssl.conf.template`. Other branches and `frame-ancestors` unchanged.
  
  ### Testing
  `nginx -t` passes on the rendered template; envsubst leaves `blob:` and `${FRAME_ANCESTORS}` intact. Verified end-to-end against a locally built gateway: the live `/openmrs/spa/` response carries `object-src 'self' blob:`, and a blob `<object>` PDF loads in Chromium with zero `securitypolicyviolation` events — the same page is blocked under `main`'s CSP.

### DIFF
--- a/gateway/default-ssl.conf.template
+++ b/gateway/default-ssl.conf.template
@@ -1,5 +1,5 @@
 map $request_uri $csp_header {
-  default "default-src 'self' 'unsafe-inline' 'unsafe-eval' localhost localhost:*; base-uri 'self'; font-src 'self'; img-src 'self' data:; frame-src 'self' blob:; frame-ancestors 'self' ${FRAME_ANCESTORS};";
+  default "default-src 'self' 'unsafe-inline' 'unsafe-eval' localhost localhost:*; base-uri 'self'; font-src 'self'; img-src 'self' data:; object-src 'self' blob:; frame-ancestors 'self' ${FRAME_ANCESTORS};";
   "~^/openmrs/(?:admin|dictionary|module|patientDashboard.form)/" "default-src 'self' 'unsafe-inline'; script-src 'self' 'unsafe-inline' 'unsafe-eval'; base-uri 'self'; font-src 'self'; frame-ancestors 'self';";
   "~^/openmrs/owa" "default-src 'self' 'unsafe-inline'; script-src 'self' 'unsafe-inline' 'unsafe-eval'; base-uri 'self'; font-src 'self' data:; img-src 'self' data:; frame-ancestors 'self';";
 }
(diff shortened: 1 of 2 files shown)


BlessedAmrita:
cc: @wikumChamith @NethmiRodrigo

ibacher:
Is there a reason we can't use the same technique we did to display PDF attachments?

BlessedAmrita:
Update on the preview CSP issue. My #1059 targets the wrong directive: the preview uses a blob <object>, governed by object-src, not frame-src, so under #1059 it's still blocked (confirmed in Chromium against the real gateway CSP). Ian's question helped catch it!

Two fixes, both built and CSP-tested:

A) Same-origin URL like attachments. Loads clean under main's current CSP, zero violations, so no distro change and #1059 closes. Cost: the browser owns the fetch so we lose the status-based error UX, including the 404 "your module predates the endpoint" diagnostic (4 tests dropped).

B) Keep the blob, swap <object> to <iframe>. Needs the distro change but correctly this time, and keeps all 18 tests and the full error taxonomy.

So: A keeps the gateway CSP untouched but loses the 404 diagnostic, B keeps every diagnostic but needs the CSP change. I lean A, but the module-version 404 is worth weighing. What do you all think...?

cc @wikumChamith @ibacher @NethmiRodrigo

wikumChamith:
> Update on the preview CSP issue. My #1059 targets the wrong directive: the preview uses a blob , governed by object-src, not frame-src, so under #1059 it's still blocked (confirmed in Chromium against the real gateway CSP). Ian's question helped catch it!
> 
> Two fixes, both built and CSP-tested:
> 
> A) Same-origin URL like attachments. Loads clean under main's current CSP, zero violations, so no distro change and #1059 closes. Cost: the browser owns the fetch so we lose the status-based error UX, including the 404 "your module predates the endpoint" diagnostic (4 tests dropped).
> 
> B) Keep the blob, swap to <iframe>. Needs the distro change but correctly this time, and keeps all 18 tests and the full error taxonomy.
> 
> So: A keeps the gateway CSP untouched but loses the 404 diagnostic, B keeps every diagnostic but needs the CSP change. I lean A, but the module-version 404 is worth weighing. What do you all think...?
> 
> cc @wikumChamith @ibacher @NethmiRodrigo

@ibacher any thoughts on this?

ibacher:
I think I'm ok with this PR going ahead, but we do need to change the rule.

BlessedAmrita:
Thanks @ibacher! Updated the rule to object-src 'self' blob: since the preview uses a blob `<object>`, and confirmed against the live gateway in Chromium that the `<object>` now loads with zero violations (object-src alone is sufficient, so I kept it minimal). Both templates, the commit, and the PR description updated.

denniskigen:
@ibacher mind re-reviewing?

ibacher (approved):

PR #1069, opened 2026-08-31
O3-5937: Collect and visualize JVM metrics from the OTel pipeline

jayasanka-sack:
### What

The backend exports JVM metrics in OpenTelemetry format. this PR is the other half: picking them up and making them useful.

Alloy receives OTLP, hands off to Prometheus, and a new dashboard renders it. also tidied the dashboards into monitoring/grafana/dashboards/ since they were sitting loose in monitoring/.

### Ticket:
https://openmrs.atlassian.net/browse/O3-5937

Blocked by: https://github.com/openmrs/openmrs-core/pull/6522

### Why

We've had no visibility into JVM health. no heap, no GC, no thread pools. so "is the server about to fall over" was guesswork.

The export side speaks standard OTel, which means this collection setup is one option rather than the option. Implementers on Datadog, Honeycomb, or their own collector can point the backend somewhere else entirely and this PR becomes irrelevant to them. that's the point:

> ### _OpenMRS exports standard OTel. Grab it however you like. or use our setup, if you want 👀_

### Screenshots
image


docker-compose.grafana.yml
```diff
@@ -77,6 +83,21 @@ services:
             monitoring-init:
                 condition: service_completed_successfully
 
+    backend:
+      environment:
+        OMRS_OTEL_ENABLED: true
+        OTEL_EXPORTER_OTLP_ENDPOINT: http://alloy:4317
+        # Set explicitly: the agent defaults to http/protobuf, which cannot talk
+        # to the gRPC listener on 4317. Use http/protobuf + :4318 instead if your
+        # collector is only reachable over HTTP/1.1.
```
ibacher, line 92: Is this helpful to have documented here? If so, I think this comment needs to be reworded, e.g.,


```suggestion
        # Default is Protobuf over HTTP, we're using the gRPC listener instead because...
        # Protobuf over HTTP runs on port 4318 by default.
```

docker-compose.grafana.yml
```diff
@@ -77,6 +83,21 @@ services:
...
+        OTEL_EXPORTER_OTLP_PROTOCOL: grpc
+        OTEL_SERVICE_NAME: openmrs-backend
+        OTEL_JMX_TARGET_SYSTEM: tomcat
+        OTEL_METRICS_EXPORTER: otlp
+        OTEL_TRACES_EXPORTER: none
+        OTEL_LOGS_EXPORTER: none
+        OTEL_RESOURCE_ATTRIBUTES: "service.namespace=openmrs,deployment.environment.name=dev"
```
ibacher, line 99: This feels like maybe it should be commented. Is a deployment environment name required? Because it's easier if we aren't forced to guess.

```suggestion
        OTEL_RESOURCE_ATTRIBUTES: "service.namespace=openmrs"
```

README.md
```diff
@@ -232,14 +232,78 @@ The certbot entrypoint skips certificate generation when it finds existing certi
...
+local volume. For production or multi-replica deployments, reuse the configuration pattern
+shown here rather than the compose file itself.
+
+### Backend metrics (OpenTelemetry)
+
+JVM metrics come from the OpenTelemetry Java agent, which is bundled in the `openmrs-core`
+base image (downloaded and checksum-verified in that image's Dockerfile). Setting
+`OMRS_OTEL_ENABLED=true` makes the backend's startup script attach the agent to Tomcat.
+No OpenMRS code reads the `OTEL_*` variables -- the agent picks them up from the
+environment itself, so this compose file is the entire configuration surface.
```
ibacher:
```suggestion
```

README.md
```diff
@@ -232,14 +232,78 @@ The certbot entrypoint skips certificate generation when it finds existing certi
...
+
+The path is: Java agent -> OTLP/HTTP -> Alloy -> Prometheus -> Grafana.
+
+Because the agent auto-instruments the whole web application, it exports HTTP server and
+JDBC client metrics in addition to the `jvm.*` family. The provisioned dashboard plots the
+JVM metrics only; anything else the agent sends is still queryable in Prometheus.
+
+#### Configuration options
+
+| Variable | Agent default | Set by this stack | Description |
```
ibacher: "Agent default" is awkward for this column

README.md
```diff
@@ -232,14 +232,78 @@ The certbot entrypoint skips certificate generation when it finds existing certi
...
+|----------|---------------|-------------------|-------------|
```
ibacher: This sort of overlaps or at least the document doesn't clearly distinguish these settings from the existing Environment Variables Reference.

ibacher (commented):
I think the code is fine, but we need to clean-up the documentation a bit.

README.md
jayasanka-sack: Just to make sure I've got this right. You mean these end up being maintained in two places so they can drift apart?

If that's the case, I think moving them into the compose file would help.

jayasanka-sack: fixed with 6b74d4d68c6e94e29d97da72d1963d28e2ee3795

ibacher (approved):

PR #1077, opened 2026-09-21
(feat):  Add SPA_ENV variable to configure build environment

NethmiRodrigo:


NethmiRodrigo:
@ibacher there's no point in merging this until I update the tooling to actually read this right?

ibacher (approved):
There’s no reason it’s harmful either, I think…

frontend/spa-build-config.json
```diff
@@ -3,6 +3,7 @@
   "apiUrl": "$API_URL",
   "configUrls": ["$SPA_CONFIG_URLS"],
   "defaultLocale": "$SPA_DEFAULT_LOCALE",
+  "spaEnv": "$SPA_ENV",
```
NethmiRodrigo: @ibacher its correct to use a new variable here and not `env` right?

ibacher: No, I think this should just be the env variable. We've never used that for anything meaningful, so it should be safe.

